PR #14, opened 2015-07-21
loader: add support for EMB_SDA21 relocations

zhuowei:
This relocation format is used in RPXes, and can be found, for example, in version 24 of Amiibo Settings.
I don't know where Cafe OS actually populates r0, r2, and r13; I populate it at thread creation.

### DIFF
--- a/src/elf.h
+++ b/src/elf.h
@@ -179,6 +179,7 @@ enum // r_info & 0xff
    R_PPC_GOT_DTPREL16_HA = 94,
    R_PPC_TLSGD = 95,
    R_PPC_TLSLD = 96,
+   R_PPC_EMB_SDA21 = 109,
    R_PPC_REL16 = 249,
    R_PPC_REL16_LO = 250,
    R_PPC_REL16_HI = 251,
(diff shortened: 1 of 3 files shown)


brett19:
Nice!  Thanks, looks like we will need to provide population of those registers on thread creation, but this is a great start.

exjam:
Hi zhuowei,

So I read up on SDA relocations from the following manuals:
http://cache.freescale.com/files/32bit/doc/ref_manual/E500ABIUG.pdf
http://www.freescale.com/files/32bit/doc/app_note/PPCEABI.pdf

From reading these pdf it would seem your code needed a few changes:
1. sdaBase is supposed to be an address such that within a 16 bit signed offset the entire .sdata and .sbss sections are accessible
2. .sdata0 sections use offsets relative to memory address 0, so you do not need to set gpr[0] to the base of .sdata0

I committed these changes but have no way to verify the behaviour, is "Amiibo Settings" a system rpx or from a game?
https://github.com/exjam/wiiu-emu/commit/88d4d23bf961fd49c2e74e96635bf83f5d24ad37

Thanks for the pull request, any future ones are always welcome! I would request that you maintain the coding style of the file you edit though. Specifically no single line if statements with no braces!

Issue #4, opened 2015-07-01
Add proper callback support from host to game code


exjam:
Ensure interpreter / jit is reentrant.
Ensure it doesn't massacre ThreadState.

PR #15, opened 2015-07-21
zlib125: implement inflateInit_, ignore custom allocators instead of asserting

zhuowei:
Some applications such as Amiibo Settings v24 uses inflateInit_ instead of inflateInit2, and does set the zalloc/zfree/opaque attributes on the zlib stream. This patch implements inflateInit as a wrapper for inflateInit2, and commented out the asserts that checks for no custom allocators in inflate


exjam:
I'll allow removing the assert for now as there is no proper callback support from kernel to game code (issue #4). Will open an issue to track zlib callback implementation #17

PR #16, opened 2015-07-21
Fix two comparison issues:

zhuowei:
This patch fixes two small issues:
 - filesystem misinterprets seek to offset 0 as a seek to the end
   of file
 - UserModule::findSection incorrectly returns the previous section if symbol is at
   beginning of section

I don't think the old seek behaviour is correct, because Amiibo Settings (after sort-of implementing FSReadFileAsync by copy pasting FSReadFile) gets stuck trying to decompress a zero length file if it seeked to the end

### DIFF
--- a/src/filesystem.h
+++ b/src/filesystem.h
@@ -86,7 +86,7 @@ class HostFileSystem : public FileSystem
 
       void seek(int64_t pos) override
       {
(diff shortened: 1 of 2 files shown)


exjam:
Good catch on the filesystem seek!

PR #22, opened 2015-07-28
interpreter: jit_fallback should use 64-bit mov for thread state

zhuowei:
The threadState pointer may not fit in a 32-bit register


exjam:
p.s. I am on IRC right now if you want to talk still, I was asleep last time you came on!

PR #23, opened 2015-07-29
main: fix crash when log level not specified

zhuowei:
Whoops.

For some reason the default value isn't being picked up by docopt; manually check if value is not specified and default to trace


exjam:
Yeah I've had some problems with docopt too, might have to consider replacing it. Shame, as its a nice library (and the python version works so well!).

PR #28, opened 2015-08-02
coreinit: implement OSConsoleWrite

zhuowei:
The Health and Safety Guide has a bunch of debug OSConsoleWrite calls; was surprised that this wasn't implemented already.


exjam:
Looking at coreinit.rpl it seems to take two arguments, any idea what the other argument is?

zhuowei:
probably size of buffer.

Also added a commit fixing a typo in zlib.

PR #33, opened 2015-08-29
Fix file not found bug

Kinnay:
The function never returned the file not found error, because file always points to something, so the if comparison didn't work.


brett19:
I believe what you want to use is actually file->good() which indicates if the stream was opened successfully.  If that method does not exist, we need to expose it from the underlying std::stream.

Kinnay:
I tested it, and that works too, could you add that to the main branch or do I have to create a new pull request?

PR #35, opened 2015-08-31
Fix folders not found bug

Kinnay:
For some reason, the file pointer is null if the folder doesn't exist, but it is not null if the folder exists, but the file doesn't. The previous fix for NSMBU made the functions crash if the folder doesn't exist.


brett19:
We should probably explicitly figure out this logic at some point, but for the moment, lets fix the crash.

PR #36, opened 2015-09-02
Fix single precision float bug

Kinnay:
This fixes the NSMBU crash. According to both Wiibrew and 6xx_pem.pdf single precision floats are converted to double when they are stored. In the emulator code, they're stored twice though, as if they were paired single instructions. lfs and similar instructions seem to store them into the the register the wrong way too. That will probably have to be fixed as well.

### DIFF
--- a/src/interpreter/interpreter_float.cpp
+++ b/src/interpreter/interpreter_float.cpp
@@ -102,7 +102,7 @@ template<>
 float
 getFpr<float>(ThreadState *state, unsigned fr)
 {
-   return state->fpr[fr].paired0;
+   return (float)(state->fpr[fr].value);
 }
 
 template<>
@@ -115,8 +115,7 @@ getFpr<double>(ThreadState *state, unsigned fr)
 void
 setFpr(ThreadState *state, unsigned fr, float value)
 {
-   state->fpr[fr].paired0 = value;
-   state->fpr[fr].paired1 = value;
+   state->fpr[fr].value = (double)value;
 }
 
 void


Kinnay:
Actually, I just looked at the dolphin-emu interpreter source, and it seems to handle singles the same way as this emulator. I wonder why this fixes the NSMBU crash and why the PowerPC documents mention conversion to double though.

Kinnay:
However, I did a little testing in Dolphin debug mode and it does store the values in the fpr as doubles.

brett19:
The handling of the floating point registers and a few of the more odd instructions is a bit fuzzy when reading docs.  Will need to investigate a bit more :S

exjam:
I'd definitely go with the documentation here. I wouldn't put too much weight on dolphin's interpreter code - as it probably isn't used due to their jit being so developed.

exjam:
So it depends on the HID2[PSE] flag, when it is 0 then paired single mode is disabled and all singles are converted into doubles and stored into frD as a double.

When it is set to 1 then singles are stored in both the frD.ps0, frD.ps1 as singles.

From the documentation it seems the current implementation is correct, however its weird that this change fixes a crash for you...

Kinnay:
Hm, the flag might be disabled at that point? I'm not sure how that kind of stuff works.

exjam:
https://github.com/exjam/wiiu-emu/commit/f21961cb65587b27dfad4828cc5e1c24cef2383e

phire:
Do not believe any documentation which implies paired singles stores two singles in the space of double register. 

As you have already found out, lfs/stfs always convert between a single to a double when loading and back again when storing, so a single precision number in a floating point register is converted into a double representation. This get a little complex with the handling of denormal numbers, especially with non-IEEE 754 compliant mode (which has caused many bugs in dolphin)

f21961c is more or less correct. You can load a double to paired0, then load a single to paired1 and as long as you don't use a paired single math instruction, you can recover the full double out of paired0. Also, you can load a single to paired1, do double operations in paired0 and still extract the same single out of paired1. (Single operations are defined to put their result in both paired0 and paired1 if HID2[PSE] is set)

So at paired0 is always at least 64 bits and paired1 least 33 bits (33 bits is required to cover the full range of denormal singles in double format) I believe someone has actually done a hardware test to prove that paired1 has more than 33 bits. I can't remember the exact result, but I think paired1 actually had a full 64 bits. You can test this using ps_merge00, ps_merge10 and ps_merged11 instructions to swap a double value stored in paired0 to paired1 and back.

exjam:
Thank you for the information! :+1: 

I need to revisit the FPU using dolphin as a reference in order to ensure accurate behaviour that is comparable to real hardware - but for now I just threw together something simple that worked. Will work on this once it starts causing us problems (or when we write a proper JIT).

PR #43, opened 2015-10-14
Fix some assignments in conditions.

lioncash:


brett19:
Nice catches.  Thanks @lioncash !

PR #47, opened 2015-10-14
jit: fix a comparison in shiftTo

lioncash:
Assuming this is what was actually intended (shl's branch would never actually be executed previously)


lioncash:
ping @brett19 (iirc you're working on the JIT code, based on the commit log)

exjam:
BTW the JIT is pretty neglected atm, not sure it even runs right now ;D.

PR #50, opened 2015-10-15
Normalize file flags and line endings

Tilka:


exjam:
Thanks for that, I apologise for us peasant Windows users.

PR #51, opened 2015-10-15
Interpreter: fix inputs for float multiplies

Tilka:
add/sub/div use a and b while mul uses a and c. PowerPC is weird like that.


Tilka:
Totally untested, since I'm on Linux right now.

PR #54, opened 2015-10-19
bitutils: Use memcpy instead of reinterpret_cast for bit_cast

lioncash:
Doesn't have the potential to break strict aliasing or alignment requirements this way.

### DIFF
--- a/src/bitutils.h
+++ b/src/bitutils.h
@@ -1,14 +1,21 @@
 #pragma once
 #include <cstdint>
 #include <cstdlib>
+#include <cstring>
+#include <type_traits>
 
 // reinterpret_cast for value types
 template<typename DstType, typename SrcType>
 static inline DstType
-bit_cast(SrcType src)
+bit_cast(const SrcType& src)
 {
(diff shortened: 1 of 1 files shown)


src/bitutils.h
```diff
@@ -1,14 +1,21 @@
 #pragma once
 #include <cstdint>
 #include <cstdlib>
+#include <cstring>
+#include <type_traits>
 
 // reinterpret_cast for value types
```
CryZe: reinterpret_cast :^)

exjam: The comment is still correct, as it just describes what the function does: bit_cast is the effectively the equivalent of reinterpret_cast but for value types.

reinterpret_cast lets you cast pointers:
reinterpret_cast<uint32_t*>(float* a) would cast a float pointer to a u32 pointer, if you dereference this pointer you end up with the same functionality as bit_cast

bit_cast lets you cast the values:
bit_cast<uint32_t>(float a) would cast a float to a uint32_t without type conversion, so a float 1.0f would become uint32_t 0x3F800000

PR #56, opened 2015-10-19
Add fctiw, fctiwz, frsp, fabs, fnabs, fmr, and fneg to the JIT.

MoochMcGee:
This was done with implementations ported from Dolphin.

### DIFF
--- a/src/cpu/jit/jit_float.cpp
+++ b/src/cpu/jit/jit_float.cpp
@@ -6,12 +6,84 @@ namespace cpu
 {
 namespace jit
 {
+	alignas(16) static const u64 psAbsMask[2] = { 0x7FFFFFFFFFFFFFFFULL, 0xFFFFFFFFFFFFFFFFULL };
+	alignas(16) static const u64 psSignBits[2] = { 0x8000000000000000ULL, 0x0000000000000000ULL };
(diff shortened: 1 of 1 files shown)


brett19:
I would like to run these against hardware before we merge.  I'll get @exjam to look at that.  Otherwise they look perfect :)

exjam:
Real men merge now, test later.

src/cpu/jit/jit_float.cpp
```diff
@@ -14,6 +18,74 @@ namespace jit
...
+	   a.movdqu(a.xmm0, a.ppcfpr[instr.frB]);
+	   a.movsd(a.ppcgpr[instr.frD], a.xmm0);
+
+	   return true;
+   }
+
+   static bool fsign(PPCEmuAssembler& a, Instruction instr)
+   {
+	   a.movdqu(a.xmm0, a.ppcfpr[instr.frB]);
+	   a.xorps(a.xmm0, a.ptr(psSignBits);
```
Tilka: This doesn't even compile.

exjam: Yes I will setup some CI service tomorrow haha.

PR #57, opened 2015-10-20
Removed packing alignment for structs in elf.h , since...

VelocityRa:
...it is known to degrade performance ([source](http://stackoverflow.com/questions/7793511/are-there-performance-issues-when-using-pragma-pack1)), requiring two memory accesses instead of one.
Of course we could check the performance improvements later, but if doesn't this mess up some other code(please tell me if it does), we could get a performance boost easily.

### DIFF
--- a/src/elf.h
+++ b/src/elf.h
@@ -7,8 +7,6 @@ class BigEndianView;
 namespace elf
 {
 
-#pragma pack(push, 1)
-
 enum // e_machine
 {
    EM_PPC = 20 // PowerPC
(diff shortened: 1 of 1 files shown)


VelocityRa:
Just did a Grep search and I noticed that this is in a lot of other files as well...

exjam:
Pretty much everywhere there is pragma pack it is there for a reason.

pragma pack means the compiler puts no padding into structures - this is very important when we are trying to duplicate the exact memory layout of structures on the Wii U.

However, this place in elf.h is probably the only place it is not needed (at the moment) due to how we have set up the file reading, although I feel we should change the elf structs to use be_val and remove the dependency on BigEndianView and consequently we will need to keep pragma pack.

VelocityRa:
Ah, I understand, thanks for explaining.

PR #63, opened 2015-10-22
bitutils: Add Linux and Mac byteswap implementations

archshift:
_byteswap_u*** is a MSVC intrinsic and is non-portable.

### DIFF
--- a/src/utils/bitutils.h
+++ b/src/utils/bitutils.h
@@ -5,6 +5,10 @@
 #include <cstring>
 #include <type_traits>
 
+#ifdef __linux__
+#include <byteswap.h>
+#endif
+
 // reinterpret_cast for value types
 template<typename DstType, typename SrcType>
 static inline DstType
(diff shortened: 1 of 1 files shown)


src/utils/bitutils.h
```diff
@@ -211,7 +215,15 @@ struct byte_swap_t<Type, 2>
 {
    static Type swap(Type src)
    {
+#ifdef _MSC_VER
       return bit_cast<Type>(_byteswap_ushort(bit_cast<uint16_t>(src)));
+#elif __APPLE__
+      // Apple has no 16-bit byteswap intrinsic
```
lioncash: OSSwapInt16

archshift: Not exactly an intrinsic!

lioncash: Right, but why not use it to reduce redundancy, since it still does the swap.

brett19:
While I agree with this commit in practice, I think we should move platform specific stuff out to its own file rather than polluting arbitrary files with platform-specific things.

archshift:
It's a tradeoff between organization and redundancy, I guess.

You can make platform specific byteswap files, which would change this single-file implementation to be four+ different files each with maybe three function definitions each.
You can make some kind of common_win, common_mac, common_linux etc system, but you'd lose the organization that you already have here with a file with a specific purpose.
You can just have the ifdefs where they would locally be needed, where you'd lose some organizational capacity (but not much!) but keep everything simple and manageable.

exjam:
I think a better compromise would be to make platform.h have defines for each platform like below which we can use rather than the platform specific defines like __linux__ which are harder to remember.
PLATFORM_WINDOWS
PLATFORM_LINUX
PLATFORM_APPLE

I agree that files should be split by functionality, rather than alllll platform specific stuff in one file. For filesystem I ended up making a header like filesystem_host_folder.h which uses ifdefs to include the platform specific implementation filesystem_win_folder.h.

The current platform.h is pretty much all window creation, so really that should be refactored to ui/window.h or something. It has set thread name in it too which should be put into its own file.

PR #65, opened 2015-10-22
Fix initializer list order

lioncash:
Will stifle eventual warnings on compilers other than MSVC.

### DIFF
--- a/src/mem/mem.cpp
+++ b/src/mem/mem.cpp
@@ -31,12 +31,12 @@ enum class MemoryType
 struct MemoryView
 {
    MemoryView() :
-      address(0)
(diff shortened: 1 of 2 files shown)


brett19:
The first commit here seems odd, I'm not aware of any particular enforcement of initializer list order?  Especially consider when a initializer item needs data from a previous initializer item.

lioncash:
Classes initialize in the order their members are declared in. This is in the standard.

```cpp
class SomeClass
{
public
    SomeClass() : y(3), x(9) {}

private:
    int x;
    int y;
}
```

will still result in x being initialized before y.

PR #67, opened 2015-10-23
trace.h: don't declare TraceFieldValue::ValueType in anonymous union

archshift:
The CPP standard forbids named struct declarations in anonymous unions, MSVC allows it however.

### DIFF
--- a/src/cpu/trace.h
+++ b/src/cpu/trace.h
@@ -32,6 +32,16 @@ namespace StateField {
 typedef uint32_t TraceFieldType;
 
 struct TraceFieldValue {
+   struct ValueType {
+      bool operator==(const ValueType& rhs) {
+         return data[0] == rhs.data[0] && data[1] == rhs.data[1];
+      }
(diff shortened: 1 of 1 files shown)


brett19:
I assume you mean the standard forbids non-pod struct declarations?

archshift:
Nope, see [this LLVM example](https://llvm.org/svn/llvm-project/cfe/trunk/test/SemaCXX/anonymous-union.cpp), specifically:
```cpp
struct BadMembers {
  union {
    struct X { }; // expected-error {{types cannot be declared in an anonymous union}}
  }
}
```

I probably should have written that it forbids **named** struct declarations in anonymous unions.

This follows the C++ standard section 9.5.5:
> Nested types, anonymous unions, and functions cannot be declared within
an anonymous union

brett19:
Fair enough.  We are going to be refactoring the tracing soon so that PSR0 and PSR1 are split instead of stored together to reduce the size of the trace data, and it will remove the need for this stupid-little-thingy™.

PR #89, opened 2015-11-02
coreinit_debug: Use std::array instead of std::vector for fixed tables

lioncash:


exjam:
It's effort having to define the size - if we want a std array here might be better to create a make_array function which returns a std::array of the correct size and use auto for the declaration.

Would probably be something like:
template<typename Type>
auto make_array(Type &&... values) -> std::array<Type, sizeof...(Type)>
{
return {{ std::forward<Type>(values)... }};
}

lioncash:
@exjam Your template arguments are cut off (need to escape the \< and \>)

exjam:
Fixed, also did a quick google and found a std proposal but its a bit over engineered.

http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n3824.htm
https://gist.github.com/lichray/6034753

Can't test anything myself right now, so no clue if my solution works :+1:

PR #90, opened 2015-11-02
Change some zero literals to nullptr

lioncash:
If wanted, I can also change over the intentional null pointer dereference to just be std::abort or std::exit.


exjam:
Yeah probably should change the null pointer deref to gsl::fail_fast

PR #100, opened 2015-11-29
coreinit_alarm: apply not to entire if statement, not left side

archshift:
The previous functionality would be analogous to:
```
if ((!alarm->state) == OSAlarmState::Set) {
	...
}
```


lioncash:
could just use !=

archshift:
touché.

exjam:
Good catch

PR #101, opened 2015-11-30
Linux build/startup fixes

achurch:
Just enough to get the project to build and run hwtests. Hope I'm not getting in the way!

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -136,7 +136,9 @@ int main(int argc, char **argv)
       result = hwtest::runTests("tests/cpu/wiiu");
    }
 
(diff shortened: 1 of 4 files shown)


src/main.cpp
```diff
@@ -136,7 +136,9 @@ int main(int argc, char **argv)
       result = hwtest::runTests("tests/cpu/wiiu");
    }
 
+#ifdef PLATFORM_WINDOWS
    system("PAUSE");
```
archshift: @exjam Can we get rid of this altogether?

exjam: Yup.

PR #104, opened 2015-12-01
Fix wrong implementation of bit_scan_reverse() for non-Windows.

achurch:


archshift:
I must be going crazy...I could have sworn that was how I wrote it

LGTM

PR #115, opened 2015-12-06
Trivial implemetations of _Exit() and OSFatal().

achurch:
Just so my test RPX runs correctly. I didn't see an obvious source file to put these in, so I added a new one.

### DIFF
--- a/src/modules/coreinit/coreinit_exit.cpp
+++ b/src/modules/coreinit/coreinit_exit.cpp
@@ -1,0 +1,24 @@
+#include "coreinit.h"
+#include "coreinit_exit.h"
+#include "utils/log.h"
+
+void
+_Exit()
+{
+   gLog->info("Program requested exit.");
+   _exit(0);  // TODO: pass this up to the UI layer instead
+}
+
+void
+OSFatal(const char *message)
+{
+   gLog->error("Program reported a fatal error: {}", message);
+   _exit(1);  // TODO: pass this up to the UI layer instead
+}
+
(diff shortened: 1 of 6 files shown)


src/modules/coreinit/coreinit_exit.cpp
```diff
@@ -0,0 +1,24 @@
+#include "coreinit.h"
+#include "coreinit_exit.h"
+#include "utils/log.h"
+
+void
+_Exit()
```
lioncash: Functions with underscores are reserved for use in the standard library global namespace (both ones with lowercase and uppercase characters following the underscore)

achurch: True, but this follows your existing pattern of library implementation function names. Also, since this function uses C++ linkage, its true name will be mangled and thus won't collide with any standard libraries.

brett19: That would be an bad assumption though, the mangling is not guaranteed to generate a dissimilar name.

achurch: I won't say I'm actively involved in compiler development (since I'm not) but I don't know of any existing compilers that mangle names in a way that could cause collisions with library namespace, so IMO it's more useful to preserve the correlation with coreinit function names than to ensure absolute standards compliance -- particularly since this is a standalone project rather than a library, so even if a name change did become necessary it would have no external impact. But of course you can change the name if you want.

PR #117, opened 2015-12-08
Fix CPU interpreter bugs

achurch:
These fixes have all been tested against real hardware. I haven't written tests for paired singles yet, but I wanted to get this up so you could take a look at it.

This seems to have broken decaf-emu --hwtest, possibly because the test files rely on the old FPR data structure (but I haven't looked into it in depth).

For reference, the test routine I used is at: http://achurch.org/cpu-tests/ppc750cl.s


exjam:
This is a lot to go through - will take me some time, but at first glance:

I am not sure about moving back to using 2x float union with 1x double for an FPR. We have had trouble with this before, and I'm fairly confident the conclusion we came to was that both paired0 and paired1 should be stored as doubles (this is also what dolphin does).

May I ask why the reason you made this change? Did your hardware tests imply this was correct? Or were you going with the doc as I did before?

Here is what phire said on the issue: https://github.com/decaf-emu/decaf-emu/pull/36#issuecomment-147707553

achurch:
The PowerPC 750 manual is pretty explicit that FP registers are 64 bits wide, storing either a 64-bit double (regular FP instructions) or two 32-bit floats (paired-single instructions):

"1.5.2 750CL Microprocessor Instruction Set - In addition to the 32-bit single-precision and the 64-bit double-precision floating-point operands, the 750CL implements a new floating-point operand type: paired single-precision. The paired single operand uses a 64-bit FPR to maintain two 32-bit single precision floating point operands."

I'll do a quick test now to verify that that's in fact what the Espresso does, but if so, I'd be leery of using a structure that doesn't match the actual register layout.

achurch:
The comment you linked doesn't seem to apply to the Espresso:
```
    lfd %f30,288(%r31)  # loads 0x3FF55555_55555555 (1.333...)
    lfd %f31,304(%r31)  # loads 0x3FFFFFFF_FFFFFFFF (2.0 - 1ulp)
    frsp %f31,%f31      # force f31 to 2.0f for ps_merge00 (maybe not needed?)
    ps_merge00 %f30,%f30,%f31 # copy f31(ps0) to f30(ps1)
    stfd %f30,0(%r6)    # stores 0x3FF55555_60000000 (rounded to float!)
```

degasus:
frsp rounds to single precision

achurch:
Yes, frsp on %f31, not on %f30.

achurch:
Just for the record, I will be going through the 750CL instruction list and writing tests for all the paired-single instructions. It may end up being more convenient to have "double paired[2]" with paired[0] overlaying the non-PS 64-bit value. But at the least, the previous layout was broken with respect to non-PS instructions.

exjam:
It is possible the Wii and Wii U processors differ on this - I will have to run some hardware tests to figure it out

degasus:
What will happen on:
```
lfs %f31
lfd %f31
ps_merge11 %f31, %f31, %f31
stfd %f31
```

achurch:
Looks like the high 32 bits of the loaded double get interpreted as a float:
```
lfs %f31,4(%r30)    # loads 0x3F800000
lfd %f31,288(%r31)  # loads 0x3FF55555_55555555
ps_merge11 %f31,%f31,%f31
stfd %f31,0(%r6)    # stores 0x3FFEAAAA_A0000000
```
I'm working on load/store tests now, will make concrete suggestions once I have a clearer picture of how PS and normal instructions interact.

exjam:
Yeah I'm going to work on getting some coverage of paired instructions and load/store using my auto generated hwtests. Not sure how much they will help for this though as they only test 1 instruction at a time.

degasus:
@achurch By the way, your first example did already proove that this union is wrong. To get this result, you need at least one additional bit for storing how to reinterpretate the data within this 64bit register.

achurch:
I agree, and the "hidden bit" is my current working model. As I said, I'll make more concrete suggestions once I've written more tests.

degasus:
Do you also test to load two doubles which are a bit too big for storing them as a float. Afterwards call fsubs on them? Is the result finite again?

achurch:
That will always be finite because fsubs performs its operation in double precision and only rounds the final result.

degasus:
Oh, did they fix the documentation about that? On the wii, the specs did say that the precision is fixed also on inputs.

achurch:
I don't know about the Wii documentation, but the 750CL manual explicitly says "the result is rounded to single-precision". I just added a couple of tests and confirmed that's what the Espresso does.

achurch:
Still need to look at fmr/fneg/etc and the FPSCR instructions, but here's the tentative fix to deal with paired singles. Will sleep on it and see if I find any more cases to worry about tomorrow.

achurch:
@degasus: Regarding fsubs, do you have any specific examples of cases that seemed to be performed in single precision? It turns out the PowerPC UISA has an explicit note saying single-precision operations on operands outside the single-precision range are undefined, so the CPU might well behave differently depending on the operand values. (I ran the same 8 tests I used here on a PS3, and 3 of them failed with NaNs even though all 8 are doing more or less the same computation.)

degasus:
I think fiora described that very well at https://www.alchemistowl.org/pocorgtfo/pocorgtfo06.pdf on page 8 - 3.3. Maybe the expresso still have the same FPU?

achurch:
Could be, though that's a different issue (precision vs. range). I'll write tests for precision as well, though at the moment I'm still focusing on paired singles.

exjam:
I just tested
   lfs %f2, 0x0(%r5)   (load single value 32.0f into f2, which effectively sets fpr, ps0 and ps1)
   lfd %f2, 0x0(%r3)   (load value 64.0 into f2)
   fadd %f2, %f2, %f1  (add value 1.0 onto 64.0 and store into f2)
   psq_st %f2, 0x0(%r3), 0, 0  (paired single store of f2 into r3)

The output was in 0x0(%r3): 4282000042000000, which is a 32 bit single 65.0f and a 32 bit single 32.0f.

So this confirms that ps0 is the same than the fpr value, and ps1 is a separate value entirely. Thus the registers should not be a union. I think we should have ps0 and fpr be the same double, and a separate double for ps1, both of which round to single when necessary such as on a psq_st, or a stfs.

achurch:
I was just about to write a comment saying I've found the same thing (:
So far it looks like ps1 should be a float, though. Need to do deeper testing.

achurch:
Also I have a feeling the relationship between lfd and ps1 is not stable; depending on the instruction sequence, sometimes it writes the upper 32 bits of the double into ps1 (as I posted earlier) and sometimes it leaves ps1 alone. I may set that aside for the moment and get back to it after I look into the rest of the ps instructions.

degasus:
What will happen if you emit lots of NOPs between those calls? Maybe the pipeline does some unsafe optimizations....

achurch:
I tried that but couldn't find any difference due to timing. I think it may be more an issue of data hazards caused by particular instructions being near each other.

exjam:
With using psq_l to load the register gives a value which is to be expected if ps0==fpr.
   psq_l %f2, 0x0(%r3), 0, 0   (load paired 32.0f, 32.0f into ps0, ps1)
   fadd %f2, %f2, %f1             (f2 + 1.0f)
   stfd %f2, 0x0(%r3)              (store f2)

0x0(%r3) = 65.0f

Maybe it depends on whether you use floating operations or ps_ operations on the register, or just ps_merge could be a whole edge case on its own?

achurch:
That's odd, how'd it give 32 + 1 = 65?
At any rate, my current theory is that each FPR has a shadow 32-bit register which functions as ps1 for the paired single instructions, and ps0 is just read from the base 64-bit FPR. I've found that lfd -> ps_cmpu0 compares the full 64-bit value (i.e. equality tests fail if you use either a truncated or a rounded value), though I haven't gotten to the actual arithmetic insns yet.
I did see that ps_merge _writes_ float values even if reading from a double (so e.g. lfd 0x3FF55555_55555555 -> ps_merge -> stfd gives 0x3FF55555_60000000), and to make things even more fun, ps_merge rounds ps0 but truncates ps1 (0x3FF55555_40000000 in this case).

exjam:
My bad meant to type 64.0f. Yeah seems like ps0,fpr are the same 64 bit register and ps1 is a separate 32 bit register.

achurch:
Okay, I've confirmed ps_add also sees the full precision of a loaded double value, so I've gone ahead and pushed the structure change adding a separate ps1 field.

exjam:
Would it make sense to make paired1 also a double and do rounding when needed as it would make JIT code easier if both are doubles?

brett19:
Aka: Back to where we started, no?

exjam:
For the register structure at least, but there are many FPU fixes here!

achurch:
For my part I'd rather (at least for the moment) keep the register structure close to what the hardware seems to be actually doing. I agree it may be easier for JIT to have them both the same type, but I'd rather not think about optimizing until after I'm done with correctness. (:

achurch:
On the topic of PSQ load/store, do you know if there's a way to set GQRs on the Wii U? It looked at first as though I could just mtspr to them, but apparently that only works if I load zero into the register, and any other value (e.g. 0x00040000 to load unsigned bytes) causes the system to crash, presumably with a privilege exception.

exjam:
You have to set UGQR (user level GQR), not GQR (supervisor level). They have different ID to normal GQR.

You can see them in code here:
https://github.com/decaf-emu/decaf-emu/blob/master/src/cpu/state.h#L330-L337

Which is copied from this wiki:
http://wiiubrew.org/wiki/SPRs

achurch:
Ah, the 750CL manual doesn't mention UGQRs. Maybe they're specific to the Espresso?
Thanks.

exjam:
Yeah I'm fairly sure they're specific to Espresso, games run strictly in user space but some supervisor level registers like GQR could still be useful to games so I guess that's why they mapped them with user level IDs too.

achurch:
I still need to look at the scalar/PS instruction interactions, but this should at least cover everything documented in the 750CL manual.

exjam:
Looking good! Not sure if you ever ran hwtests but:

Before your changes we had Passed: 65600, Failed: 5696, all of which were in FMA: fmadds, fmsub, fmsubs, fnmadd, fnmadds, fnmsub, fnmsubs.

Now we get Passed: 71256, Failed: 40, so we still have a few cases where we are not 100% the same than Wii U CPU with the single FMA instructions (fmadds, fmsubs, fnmadds, fnmsubs). Awesome progress!

If you're interested in what failed, here is the log:
https://gist.github.com/exjam/eac03f53eb803881999f

achurch:
Thanks for the merge! Interesting list of failures, e.g. -1 * HUGE_VAL + HUGE_VAL = -inf; I'll check those out when I look into that multiplier detail from PoC||GTFO.

exjam:
BTW there are 4 places where you pass a double to expand_float, I wasn't sure if just adding a static_cast<float> just to remove the warnings is the behaviour you intended.

1>interpreter_float.cpp(259): warning C4244: 'argument': conversion from 'double' to 'float', possible loss of data
1>interpreter_float.cpp(260): warning C4244: 'argument': conversion from 'double' to 'float', possible loss of data
1>interpreter_float.cpp(470): warning C4244: 'argument': conversion from 'double' to 'float', possible loss of data
1>interpreter_float.cpp(471): warning C4244: 'argument': conversion from 'double' to 'float', possible loss of data

         state->fpr[instr.frD].paired0 = extend_float(d);
         state->fpr[instr.frD].paired1 = extend_float(d);

exjam:
Also the list of values that the FPR tests execute are:

{
   0.0,
   -0.0,
   1.0,
   -1.0,
   std::numeric_limits<double>::min(),
   std::numeric_limits<double>::max(),
   std::numeric_limits<double>::lowest(),
   std::numeric_limits<double>::infinity(),
   -std::numeric_limits<double>::infinity(),
   std::numeric_limits<double>::quiet_NaN(),
   std::numeric_limits<double>::signaling_NaN(),
   std::numeric_limits<double>::denorm_min(),
   std::numeric_limits<double>::epsilon(),
   33525.78
};

So "HUGE_VAL" is ::max which "returns the largest finite value of the given type"

achurch:
Whoops, good catch! That should have been extend_float(static_cast<float>(d)) -- note that I use extend_float rather than static_cast<double> to handle NaNs without raising host FPU exceptions. Some of that code may change once I figure out what's going on with fmadds; at first guess it's using a shorter intermediate result buffer than for doubles. I think I saw something in the 750CL manual about exactly how that's implemented so I'll look over that again.

PR #118, opened 2015-12-13
Fix floating-point multiplication.

achurch:
I think this takes care of all the remaining edge cases with FP multiply operations; at least, hwtests all pass now, and the logic I ended up with seems to describe a sensible hardware implementation.


exjam:
Nice one! Good job, thanks for the help!

The only remaining thing for a 100% accurate FPU implementation is full FPSCR emulation, I'm not sure I want to pursue implementation of that as it sounds like a real pain and I reckon our code would end up super messy. If you are interested to see how all the tests fail you can set TEST_FPSCR = true at the top of hwtests.cpp, note that all of the FPU tests in hwtests start with fpscr fully zeroed. So we still do not test the multiple rounding modes, ieee mode etc.

brett19:
:-1: You should keep the fm___ instructions as frA, frB, frC in the instruction table.  This is the way they are presented in the PPC manual, and the instructions are frA * frC + frB.

achurch:
@brett19: Uh, nope. The PPC manual lists them as frA, frC, frB.

achurch:
@exjam: My own tests already check FPSCR with the exception of FR (which I agree it's probably not worth implementing, since it would mean we'd have to implement most FP operations in software). It looks like 343 hwtests fail if I enable checking FPSCR with FR excluded; I'll take a look at those just for completeness.

achurch:
@brett19: To follow up, I did come across one "Enhanced PowerPC" manual that listed "fmadd frT, frA, frB, frC" in the instruction index, though even that manual uses "frA, frC, frB" in the actual instruction description and the sample code. But both the 750CL manual and every PowerPC ISA manual I've been able to find (even the first edition, from 1993!) consistently list the operands in "frA, frC, frB" order. So I suspect that if you have one listing "frA, frB, frC" then it's a typo.

PR #120, opened 2015-12-15
Fix FPSCR behavior.

achurch:
It turns out this was fairly non-intrusive; the only real additional complexity is VXISI for fmadd/etc and over/underflow on fres. I was actually able to simplify roundForMultiply() once I discovered that one of the causes of test failures was a bug in Intel's FPU (at least on the Haswell chip I have) causing FMA3 instructions to not raise the underflow exception in certain cases when they should.

There's also one minor result-affecting fix, which is that ppu_estimate_reciprocal() accepted too narrow an exponent range.

Ignoring FPSCR[FR], there are 26 hwtests that still fail. 2 of those are in fres, in which we don't set FPSCR[FI] on a non-over/underflow inaccurate result (to be fair, the manual says that fres leaves FI undefined). The other 24 are in fmsub and fnmsub and are caused by the Intel FPU bug; fmadd and fnmadd have the same issue, but none of the operand sets in the hwtests trigger the bug for those instructions. I doubt either problem will have any impact on real software.

### DIFF
--- a/src/cpu/interpreter/interpreter_float.cpp
+++ b/src/cpu/interpreter/interpreter_float.cpp
@@ -519,9 +519,10 @@ fmaGeneric(ThreadState *state, Instruction instr)
    const double addend = (flags & FMASubtract) ? -b : b;
 
    const bool vxsnan = is_signalling_nan(a) || is_signalling_nan(b) || is_signalling_nan(c);
-   const bool vxisi = ((is_infinity(a) || is_infinity(c)) && is_infinity(b)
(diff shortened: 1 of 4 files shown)


achurch:
As noted in that last commit, the underflow thing was not actually an FPU bug but a difference in design which IEEE 754 happens to allow. I still doubt it will be an issue for real software, though if needed it could be worked around by detecting a result of magnitude std::numeric_limts::min() and rerunning the operation in round-toward-zero mode.

exjam:
Awesome progress, thanks!

PR #122, opened 2015-12-18
Fix CMake build.

achurch:
Also noticed some stray executable bits on source files and cleared those out.


exjam:
Thanks for this, I was being lazy.  And yeah whenever I create a new file it ends up with +X on it, not sure why :(.

achurch:
Apparently Windows just defaults everything to +x: http://superuser.com/questions/419877/ntfs-3g-using-user-mapping-all-files-become-executable-after-boot-to-windows
I don't work with Windows much, so I don't know if there's an easy way to change that or if you'd have to manually fix the permissions on each new file.

PR #127, opened 2015-12-21
Calculate proper display buffer sizes in GX2Calc{TV,DRC}Size().

achurch:
I've verified the return values for TV (at 720p) and DRC with R8_G8_B8_A8 format, so in particular, 864 instead of 854 is correct for the gamepad screen. I didn't check which formats are accepted by the hardware (mostly because I haven't been able to successfully set up a drawing environment yet), so I just wrote a generic function to return bits per pixel for all surface formats; that could probably be moved to some common source file if there are any other functions which need bits/pixel info.

### DIFF
--- a/src/modules/gx2/gx2_display.cpp
+++ b/src/modules/gx2/gx2_display.cpp
@@ -1,96 +1,6 @@
 #include "gx2_display.h"
+#include "gx2_format.h"
 #include "gpu/driver.h"
-
-static int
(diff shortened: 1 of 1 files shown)


exjam:
There is already a function for getting the BPP of a surface, in gx2_format, the data table it uses was dumped from the real gx2.rpl too.

https://github.com/decaf-emu/decaf-emu/blob/master/src/modules/gx2/gx2_format.cpp#L212-L222

achurch:
Fixed, thanks for the pointer.

PR #128, opened 2015-12-21
Implement an SDL2-based UI, and use it in CMake builds.

achurch:
I've tested keyboard and joystick input; I haven't tested graphics beyond opening the window, but that's mostly OpenGL (I think?) so I don't expect this to break anything.

If this is accepted, my intent is to next add support for separate TV and DRC windows, so you could (for example) make the TV window fullscreen on one monitor and have the gamepad window on another monitor. I can also set up a sound output backend, as I see a bit of work is going on in snd_core.

I haven't touched the Visual Studio build since I don't have VS. Hopefully SDL shouldn't be too difficult to pull in there.

### DIFF
--- a/src/platform/platform_sdl_input.cpp
+++ b/src/platform/platform_sdl_input.cpp
@@ -47,7 +47,7 @@ static int keyboardStateLen;
 ControllerHandle
 getControllerHandle(const std::string &name)
 {
-   if (name.compare("keyboar") == 0) {
+   if (name.compare("keyboard") == 0) {
       return keyboardHandle;
    }
 
--- a/src/platform/platform_sdl_input.cpp
+++ b/src/platform/platform_sdl_input.cpp
@@ -1,0 +1,115 @@
+#include "platform_sdl.h"
+#include "platform_input.h"
+#include "utils/log.h"
+#include <SDL.h>
+#include <SDL_gamecontroller.h>
+
+namespace platform
+{
+
+namespace sdl {
+
+void
+handleEvent(const SDL_Event *event)
+{
+   switch (event->type) {
+   case SDL_QUIT:
+      shouldQuit = true;
+      break;
+   }
+}
+
+} // namespace sdl
+
+namespace input
+{
+
+enum {
+   MAX_SDL_JOYSTICKS = 16  // Arbitrary limit.
+};
+
+static const ControllerHandle keyboardHandle = 1;
+
+static const ControllerHandle joystickHandleStart = 10;
+
+static const ControllerHandle joystickHandleEnd = 10 + MAX_SDL_JOYSTICKS;
+
+struct JoystickData
+{
+   SDL_Joystick *handle;
+};
+static JoystickData joystickData[MAX_SDL_JOYSTICKS];
+
+static const uint8_t *keyboardState;
+static int keyboardStateLen;
+
+
+ControllerHandle
+getControllerHandle(const std::string &name)
+{
+   if (name.compare("keyboar") == 0) {
+      return keyboardHandle;
+   }
+
+   for (int i = 0; i < MAX_SDL_JOYSTICKS; ++i) {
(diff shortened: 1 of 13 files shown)


linkmauve:
Note that you can already do all of that in the current GLFW backend.

exjam:
To be fair I did notice one problem with GLFW when I was setting up input config, there is no way to have a unique ID for a controller, which is a bit of a problem for loading/storing config. SDL2 seems to have a way to do this so might be a better choice.

exjam:
BTW is there any particular reason you wanted an SDL2 backend over GLFW?

achurch:
As you say, SDL2 has fairly robust controller support, including predefined mappings for a number of common controller types (so you can e.g. plug in a PS4 or XB1 controller and automatically get proper mappings for A/B/X/Y and the rest). SDL2 also provides audio output, which I don't see in GLFW -- though to be fair, this is the first time I've encountered GLFW so I'm only going from what I see in a quick skim of the online documentation.

src/platform/platform_sdl_input.cpp
```diff
@@ -0,0 +1,115 @@
...
+   }
+}
+
+::input::ButtonStatus
+getButtonStatus(ControllerHandle controller, int key)
+{
+   if (controller == keyboardHandle) {
+      if (keyboardState) {
+         const int scancode = SDL_GetScancodeFromKey(key);
+         if (scancode >= 0 && scancode < keyboardStateLen
```
VelocityRa: This is probably pretty insignificant, but [I don't think you need to check](http://www.libsdl.org/tmp/SDL/include/SDL_scancode.h) if the scancode is non-negative.

achurch: The documentation for SDL_GetScancodeFromKey() doesn't say what it returns if given an invalid input, so I check both lower and upper bounds just to be safe. I figure that given the presumably low frequency with which this function will be called, this won't have any significant impact on performance.

VelocityRa: Yeah, you're right, better safe than sorry.

src/platform/platform_sdl_input.cpp
```diff
@@ -0,0 +1,115 @@
...
+static JoystickData joystickData[MAX_SDL_JOYSTICKS];
+
+static const uint8_t *keyboardState;
+static int keyboardStateLen;
+
+
+ControllerHandle
+getControllerHandle(const std::string &name)
+{
+   if (name.compare("keyboar") == 0) {
```
Subv: Missing a 'd' in the string

src/platform/platform_sdl_input.cpp
```diff
@@ -0,0 +1,115 @@
...
+   SDL_Joystick *handle;
+};
+static JoystickData joystickData[MAX_SDL_JOYSTICKS];
+
+static const uint8_t *keyboardState;
+static int keyboardStateLen;
+
+
+ControllerHandle
+getControllerHandle(const std::string &name)
```
Subv: Is this function not used at the moment?

src/platform/platform_sdl_input.cpp
achurch: Fixed, thanks for catching that. I may have broken it while adding joystick code and not noticed since I now have decaf configured for joystick input.

exjam:
I'll merge this for now and I'll work on Windows build tomorrow, tbh I'm not that bothered whether we use GLFW or SDL2. I do kind of like how simple GLFW is, it doesn't try to do too much - just a thin layer to create a window around an OpenGL context.

Might be interesting to keep both just to ensure we keep everything unreliant on the underlying platform implementation - I guess we will see how much effort it becomes to maintain both. I do plan to make an optional Qt interface at some point too with fancy settings dialogs and such, so multiple backends are inevitable.

PR #133, opened 2015-12-23
Fix SDL build, and fix configuration defaults when GLFW is disabled.

achurch:
I saw you fixed getWindowBorderHeight(); there were just a couple other issues. I also set up config.cpp so it defaults to SDL if GLFW is disabled in the build.

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -47,7 +47,13 @@ bool debug = false;
 namespace system
 {
 
+#if defined(DECAF_GLFW)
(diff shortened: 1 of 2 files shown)


exjam:
Yeah I wasn't sure what to do about config defaults, because you can compile both SDL and GLFW whilst setting your config to use the other one. Probably should get some sort of UI setup to config inputs properly anyway, but it's so much effort especially considering we don't get any games to a playable state yet!

Thanks for fixing this!

PR #134, opened 2015-12-24
Catch invalid memory accesses in emulated code

achurch:
I've done this partly to make debugging easier, but also partly as groundwork for another change I'm looking to make which will support code that draws straight to the scan buffer (as my current test program does), by making scan buffers unwritable by default and using a write violation as an indication that the scan buffer memory needs to be copied to the host framebuffer. I am of course open to suggestions on alternate ways to accomplish that, but I figure it should be useful at least for simple experimentation.

@exjam: I don't know offhand what the Windows equivalent of ```sigaction(SIGSEGV, ...)``` is -- any chance you can take care of that? Also, I haven't dug into the details of the scheduler implementation so I'm not sure my code to stop an emulated thread is correct.

### DIFF
--- a/src/mem/mem.cpp
+++ b/src/mem/mem.cpp
@@ -1,5 +1,8 @@
+#include <gsl.h>
 #include "mem.h"
+#include "platform/platform_fiber.h"
 #include "platform/platform_memorymap.h"
+#include "processor.h"
 #include "utils/log.h"
 
 namespace mem
@@ -36,6 +39,16 @@ unmapMemory();
 
 static bool
 tryMapMemory(size_t base);
+
+// Handler for invalid memory accesses
+static platform::Fiber *handleAccessViolation(size_t address)
+{
+   if (address != 0 && (address < gMemoryBase || address >= gMemoryBase + 0x100000000ull)) {
+      return nullptr;  // Not in the emulated memory region
+   }
+   ppcaddr_t emuAddr = (address == 0) ? 0 : gsl::narrow_cast<ppcaddr_t>(address - gMemoryBase);
+   return gProcessor.handleAccessViolation(emuAddr);
+}
 
 // Attempt to map all memory regions with base
 static bool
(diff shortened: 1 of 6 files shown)


exjam:
I'm not 100% sure how I am going to detect changes in textures for the graphics drivers, I was thinking just calculating hashes on textures when they are drawn but it might be a bit slow to do on every texture every frame, the problem with page guards is that they can only guard at page granularity, so like 4096 bytes. Could be crazy and only page guard the pages which a texture overlaps and then hash the parts outside the guarded pages.

Also drawing straight to scan buffer has another problem in itself right now where we don't actually rely on the CPU memory for the scan buffer, we never upload it to the GPU.

Regarding this PR I'd like to revisit it in a few days when I have a bit more time for the Windows implementation, right now I'm split between family stuff and coding up something very cool (WiiU SDK/toolchain to compile RPLs which link to system libraries!)

achurch:
SGTM. I've actually got the scan buffer CPU memory issue dealt with locally using page guards (see my scan-buffer branch), though it turns out the scan buffers are actually swizzled and I'm still working out exactly how.

I agree textures will be more difficult since we can't assume page granularity for allocations. I'm not familiar with how the various memory regions work on the Wii U; is there a dedicated video RAM segment or is it all unified? If there are separate regions, we could potentially protect the entire VRAM region and use faults to trigger texture updates.

Definitely looking forward to that toolchain! I've got a couple of scripts which accomplish roughly the same thing with respect to coreinit.rpl, but otherwise I'm relying on lots of OSDynLoad_FindExport calls at the moment. But family time takes priority, so take it easy and happy holidays. (:

exjam:
All the memory is unified, textures, color buffers, scan buffers can all come from any memory region. However MEM1 (0xF4000000 - 0xF6000000, 32 MB) is small and fast so is normally used for color+scan buffers, but not textures, this is not a requirement though just a suggestion.

Scan buffers might also be tiled? Checkout the gx2_addrlib stuff and https://github.com/decaf-emu/addrlib it is ridiculous and not fun at all though, and I still have a bug with untiling which is to do with texture width/height not being rounded to element size properly.

exjam:
I finally got around to doing this, I have not actually tested any of this exception handling stuff yet though.

My next task is to modify the loader so we can load just a single .rpx file without messing around with making fake .xml stuff, then I will make a simple test RPX using my toolchain that does a bad memory access - and hopefully everything will work.

PR #142, opened 2016-03-03
Fixed a minor bug during Debug mode compilation.

Agorath:
- Namespace gx2:: was missing while referencing identifier GX2Texture in Debug assert.


exjam:
Ah thanks - I rarely ever compile the tools so they do get broken sometimes!

Agorath:
You're welcome - Glad I could help!

PR #143, opened 2016-03-11
Fix missing template instantiations and a couple of warnings

achurch:
I've been busy with other things, but just a couple of things I noticed when trying to build the current code. The updateFPRF() change is because recent Clang with -O2 seems to inline the double version of the function, so it's not available to the paired single instruction implementations.


exjam:
LGTM - yeah i've been rather busy lately myself too :-1: !

PR #145, opened 2016-06-03
CF: add KILL instruction

raven02:


exjam:
Looks good to me, thanks!

PR #146, opened 2016-06-03
GL: setup MIN_LOD , MAX_LOD and BIAS

raven02:


src/gpu/opengl/opengl_texture.cpp
```diff
@@ -729,12 +729,14 @@ bool GLDriver::checkActiveSamplers()
...
       gl::glSamplerParameteri(sampler.object, gl::GL_TEXTURE_COMPARE_FUNC, static_cast<gl::GLint>(depth_compare_function));
 
-      // Setup texture LOD
+      // Setup texture LOD/BIAS
       auto min_lod = sq_tex_sampler_word1.MIN_LOD();
       auto max_lod = sq_tex_sampler_word1.MAX_LOD();
       auto lod_bias = sq_tex_sampler_word1.LOD_BIAS();
 
-      // TODO: GL_TEXTURE_MIN_LOD, GL_TEXTURE_MAX_LOD, GL_TEXTURE_LOD_BIAS
+      gl::glSamplerParameteri(sampler.object, gl::GL_TEXTURE_MIN_LOD, static_cast<gl::GLint>(min_lod));
```
Subv: Are these supposed to be integers?
The OpenGL documentation for GL_TEXTURE_MIN_LOD for example says
```
 Sets the minimum level-of-detail parameter. This floating-point value limits the selection of highest resolution mipmap (lowest mipmap level). The initial value is -1000. 
```

raven02: Just fixed.

exjam:
Sorry for slow feedback, I have had a busy weekend! These fields are actually fixed point numbers so it's unfortunately not quite this simple. We need to do the reverse of this to get the correct value:

https://github.com/decaf-emu/decaf-emu/blob/2a5b4ea046cb5d74560339c91bb6bce302be28e4/src/modules/gx2/gx2_sampler.cpp#L9

https://github.com/decaf-emu/decaf-emu/blob/2a5b4ea046cb5d74560339c91bb6bce302be28e4/src/modules/gx2/gx2_sampler.cpp#L95

raven02:
Thanks .I will check this out and pull another one.

PR #148, opened 2016-06-08
Fixes compile errors on Linux

Bleizo:
Updates CMakeLists, removes some unnecessary copies and adds includes. Should fix travis-ci build.


brett19:
Hey, Thanks!

exjam:
Thanks for this, we both primarily develop on Windows where we use a VS2015 solution instead of cmake so it gets broken quite often. I tend to try fix the CMakeLists every few days though.

PR #154, opened 2016-06-27
Fix Linux build.

achurch:
It seems I only have a moment every few months to take a look at the project, but these were needed to get the current code building on Linux.

FYI, I've also submitted a fix for a race condition in glbinding that caused random crashes on startup, so you may want to update that library.


exjam:
A lot of things have changed over the past few months, good luck catching up!

Thanks for the change.

PR #155, opened 2016-06-28
Fix assertion failure when changing cores due to KC instruction.

achurch:


brett19:
This is actually not a safe fix.  When we switch cores, we should be restoring the CoreReg's state exactly as it was before the core switch.  This means that the assertion should still hold, providing that the core pointer has been updated to reference the new core.  If this assertion does not hold through a scheduler core change, it indicates that the restoral is not being performed accurately and a bug is present somewhere else.

achurch:
Fair enough. The core->cia assignment earlier in the function was explicitly listed as "for debugging purposes", so I assumed KC was an exceptional case with respect to that as well, but I'll dig deeper.

achurch:
Thanks for making me look harder. The cause of the actual case I found was not a core switch but a call to PPC code from an HLE function (specifically ghs_mtx_init), in turn called from step_one() via a kc instruction. It looks like the correct solution is to save and restore cia along with nia in wfunc_ptr::operator() - PTAL.

achurch:
Can't seem to reopen this after pushing the fixed change, so I've opened a separate request. Sorry for the update spam.

PR #156, opened 2016-06-28
Save and restore CIA across calls from native to PPC code.

achurch:
(Moved from #155 because git is being silly)


exjam:
Ah yes we didn't bother saving/restoring cia because the value wouldn't actually be used but I see how it could cause an assertion failure

PR #157, opened 2016-06-28
Remove unnecessary usage of std::bind.

Bleizo:
Because std::thread's constructors take copies of the arguments to be
used in the specified function, it is unnecessary to use std::bind to
bind the arguments of the functions to the function call.


exjam:
Thanks!

PR #161, opened 2016-07-01
Don't try to decrement past the beginning of a map.

achurch:
This crashes with a null dereference on Linux with Clang/libc++.

It looks like the test against sFuncData.end() is a no-op, but I've left it alone just in case I'm missing something.


exjam:
Makes sense :+1:

PR #165, opened 2016-07-06
Various coreinit additions.

achurch:
With respect to OSInitCoroutine/OSSwitchCoroutine, I didn't dig too deeply into the contents of the caller-saved registers other than r3-r5 (which appear to be arguments) since I imagine it's extremely unlikely any real code relies on them -- even r0 is probably just a side effect of mflr r0 in OSSwitchCoroutine itself, but that was an easy one so I included it anyway. I haven't been able to figure out what the last 0x90 bytes of the coroutine structure are for, but changing them doesn't seem to affect any user-visible registers.

I haven't actually checked the behavior of the IM* functions, but based on the names I assume they're for controlling the screensaver and auto power-off. I put the Is*Enabled functions in a header since they could potentially be used to enable/disable the host screensaver as well.


exjam:
Unfortunately coreinit coroutines are not that simple.

For one, we use a bit of an odd scheme where the current registers are actually stored in Core, and then when a thread is scheduled / unscheduled the registers are copied to / from the Core register state. @brett19 would have to come justify that decision, I know I have argued with him a lot about that, and I personally feel we should store registers in the Fiber rather than in the Core, and I can't remember what we decided at the end of that argument.

From the interpeter's point of view it is probably sufficient to add a call kernel::saveContext(thread->context) at the start of OSSwitchCoroutine, and a call to kernel::restoreContext(thread->context) at the end of it, in order to update from / to the Core registers.

saveContext & restoreContext are here:
https://github.com/decaf-emu/decaf-emu/blob/master/src/libdecaf/src/kernel/kernel_fibers.cpp#L42-L99

But this would also impact the JIT, if the calls to kernel functions such as OSSwitchCoroutine are just inlined inside a function then we aren't going to be returning to the appropriate cia/nia that OSSwitchCoroutine has set us to. This might be non-trivial, maybe we will have to make a special case in the JIT where kernel calls to specific functions which can modify the cia/nia (such as OSSwitchCoroutine) always return back to the function dispatch loop. Maybe @brett19 also has some ideas for this.

In short, did you write a test app for coroutines? Or have seen them being used in a game? If not I can write up a demo rpx using my toolchain and we can try solve this problem correctly.

exjam:
Additionally, it looks like FSGetErrorCodeForViewer takes two arguments

FSError
FSGetErrorCodeForViewer(FSClient *client, FSCmdBlock *block)

You can see that by looking at FSGetLastErrorCodeForViewer:
http://puu.sh/pSgMq/bbe548d1e7.png

Which means we'd have to store the last error code for a file command inside FSCmdBlock. It's probably worth saying that in the near future we will be updating the file system stuff to use proper guest level structures storing all the relevant information, as currently FSCmdBlock contains a std::function and FSClient uses std::vector, and we do not really want to mix host and guest memory as that can lead to memory leaks.

Funnily enough it was only recently we realised not storing the proper information in the FS structures caught us out:
https://github.com/decaf-emu/decaf-emu/commit/c8627e2df97a340ad76fbcd3a41bcb47db99fd19

The other patches look great though!

achurch:
I have a very simple test app that just calls OSSwitchCoroutine and dumps the registers, but my impetus for adding them was that Xenoblade uses them to (at least) run functions on alternate CPUs. Thanks for the info; I wasn't entirely clear on how the emulator handled context changes. I'll make the changes you suggested, and with respect to JIT, I suspect special-casing this may be the most straightforward solution, as handling context changes within JIT sounds like a headache and it doesn't seem likely they'd be a hotspot.

Regarding FSGetErrorCodeForViewer, I saw the second argument but it seemed to be dead (at least in 5.3.1) -- I traced down a couple of functions and the argument seemed to be discarded without being referenced for anything. I also thought I saw Xenoblade passing only one argument, but it looks like that was my own oversight. I'll go add the second parameter.

achurch:
@exjam, is this how you suggested modifying OSSwitchCoroutine? This changes Xenoblade's behavior significantly, in that now it seems to sit around doing nothing - though earlier it was just looping on drawing blank frames so this may be an improvement. I'll look at my test app and make sure control is getting passed correctly.

exjam:
That will probably work for the interpreter at least. I think we need to wait for @brett19 to wake up and comment on this before we proceed with it though, I imagine he will be online in a few hours.

This might be a good opportunity for me to finally add a useful HLE test application and setup proper testing scripts.

achurch:
Hmm, I checked my test app in the emulator and control is not getting passed correctly (whether I have the saveContext/restoreContext calls or not) -- it goes right through OSSwitchCoroutine as if it's a no-op. At the moment I'm copying NIA to context.lr on the assumption the native function is returning from a kc instruction to a blr, but is this not the correct way to do it?

exjam:
Not sure, later tonight I will write up a test app and play around with it, btw looks like the rest of the data is the ps1 values:
http://puu.sh/pSiON/9066a81d29.png

achurch:
Oh, interesting! Thanks for looking into that, I'll update the function.

exjam:
Also as seen from that same screenshot, gpr5 is from SPR 0x381 which is actually UGQR1.

.text:020277A0                 mfspr     r4, 0x381
.text:020277A4                 stw       r4, 8(r3)

You can look up SPR codes here:
https://github.com/decaf-emu/decaf-emu/blob/master/src/libcpu/espresso/espresso_spr.h

BTW UGQR1 is the same register as GQR1 except that it can be accessed from user level code.

achurch:
Good call, I forgot about GQRs so I didn't realize that's where that word was going.

achurch:
So the reason control wasn't getting passed correctly seems to be because I was stupid^W^Wcopied registers the wrong way when saving coroutine state. With this commit, my test app gets control passed correctly and Xenoblade now goes farther into init (eventually crashing with an assertion elsewhere - I'll look into that separately).

brett19:
Hey @achurch,
This PR is awesome!  Unfortunately the use of kernel::switchContext as you have done will likely not work properly, especially with the JIT.  It may be best if you split this into two PR's, one for the relatively sane changes and the other for the Coroutine stuff (which is likely to break things).  I will think about this more later today and see if I can find a healthy solution.
Cheers, Brett

achurch:
Got it, will do - I'll revert the coroutine stuff from here and readd it in a separate PR.

brett19:
Hey @achurch,
Could you rebase away the coroutine changes?  If you don't know how to do that I can perform the rebase for you.
Cheers, Brett

achurch:
Ah, sure - I'm not familiar with git rebase; what's the syntax?

exjam:
I will do the rebase and merge.

achurch:
Thanks, and sorry about that!

PR #168, opened 2016-07-06
gx2: Allow unaligned surface height.

achurch:
Xenoblade creates a number of surfaces on startup whose pixel count is less than MicroTileWidth * MicroTileHeight (64): for example, 8x4 R32G32F and 16x1 RGB565. For such surfaces, GX2InitColorBufferRegs computes a slice count of 0, tries to write -1U to the slice bitfield, and dies with an exception (value too large for bitfield). From a hardware design viewpoint, the most sensible behavior would be to simply round up the height to an integral multiple of the tile height when computing the slice count. But I haven't written any code to test whether this is in fact the correct behavior, so take this with a grain of salt.


exjam:
I think this is the proper fix, havn't got around to testing it on Xenoblade though.

https://github.com/decaf-emu/decaf-emu/commit/9cb621067b26e170648b16cb0c6cbc46e0833848

achurch:
Xenoblade gets through surface creation successfully with this change, so LGTM.

PR #170, opened 2016-07-10
Implement GX2DebugTagUserString functions

VelocityRa:
Kernel functions with variable arguments aren't implemented yet so I
couldn't register the one that uses them, but the functions are done.
Only game I've seen them used in is Super Meat Boy (but I do only have like 10).


VelocityRa:
Implementation is incorrect and the issue is not worth looking into any further. (the function isn't used for anything that could benefit decaf).

brett19:
For future reference, we do actually have an implementation, its just known to be incorrect in some cases and we are having trouble identifying what those cases are.

PR #172, opened 2016-07-11
coreinit: Fix handling of non-integer string format arguments.

achurch:


exjam:
Bare in mind if you see weird shit for some games, variable arguments are a bit broken.

I don't know how to decide between when we need to read the real va_list structure, and when we just read registers as normal.

PR #174, opened 2016-07-12
Debugger tweaks.

achurch:
A couple of debugger changes I found convenient, plus a workaround for being unable to select some unnamed threads.

### DIFF
--- a/src/libdecaf/src/debugger/debugger.cpp
+++ b/src/libdecaf/src/debugger/debugger.cpp
@@ -2,7 +2,9 @@
 #include "libcpu/cpu.h"
 #include "libcpu/mem.h"
 #include "debugger/debugger_branchcalc.h"
+#include "debugger/debugger_ui.h"
 #include "modules/coreinit/coreinit_internal_loader.h"
+#include "modules/coreinit/coreinit_scheduler.h"
 #include "common/decaf_assert.h"
 #include "common/log.h"
 #include <atomic>
@@ -28,9 +30,6 @@ sIsPaused;
 
 static cpu::Core *
 sCorePauseState[3];
-
-static uint32_t
-sPauseInitiatorCoreId;
 
 bool
 enabled()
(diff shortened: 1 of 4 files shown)


brett19:
👎  Do not merge.

brett19:
The logic for 8c84387 actually already exists in the code, it just doesn't work on first activation of the debugger.  It should be trivial to fix rather than needing the whole change.

brett19:
d7abed4 looks good, it's kind of strange because the hack I originally put in there (a bunch of spaces) actually lets it be selected, but then ImGui fails to actually detect clicks on it.

achurch:
Re 8c84387, where is the logic? I didn't see any, which is why the changes to add it.

brett19:
Hey @achurch, take a look at 1815f619.  I fixed the bug there.

achurch:
That commit doesn't seem to exist.

brett19:
Sorry.  Give me a few minutes.  It's on top of a bunch of other stuff I did which broke something important.

brett19:
P.S. Do you have Google Hangouts/Skype?

achurch:
Afraid not, no.

brett19:
How do you communicate with the world! :D

achurch:
E-mail. With a capital E and a hyphen. And get off my lawn! ;)

achurch:
Sure, SGTM. Thoughts on the thread highlighting commit? I found that made it easier to keep track of which thread I was looking at.

brett19:
I like the idea of that commit, I'm just not sure about how to present it, and I found the implementation a bit dirty to put it in temporarily till we can find a better solution.

achurch:
Fair enough. I'll just keep using it locally until something better comes along.

brett19:
8c84387 -> 911d8a56 (as discussed above, it just needed a minor fix)
d7abed4 -> 8bb82940 (rebased to the latest changes for you)

brett19:
040e3fc -> 7adce872 (highlight active thread)

P.S.  Hope you don't mind that I kept you as the author in spite of making minor changes.  Figured I should give credit where it is due.

achurch:
LGTM, and thanks, the credit is appreciated. (: I'm still getting some cases where the debugger doesn't stop on the right thread at a breakpoint, though -- I'll keep looking into that.
Also, nice work on the stack view! I'd actually just been thinking about making one myself.

PR #175, opened 2016-07-14
opengl: Separate shader PV/PS updates from register writes.

achurch:
Xenoblade has a shader with this ALU group:
```
   31 x: MUL             R125.x, R127.x, C6.w
      y: MUL             R125.y, R127.y, C6.w
      z: MUL             R126.z, R127.z, C6.w
      w: MULADD          R126.w, C7.z, R3.w, R126.w
      t: MULADD          R125.z, C7.z, R3.z, R126.z VEC_120
```
R126.z is both an output (in unit z) and an input (in unit t), but since these are translated to sequential GLSL statements, R126.z is clobbered before the unit t instruction has a chance to read the previous value. The operation sequence involving R126.z transforms Y coordinates from local to device space, and this bug causes a +1 on the output Y coordinates to be lost; the visible result is that a black quad intended to cover the entire framebuffer (Y=[-1,+1]) instead only covers the bottom half (Y=[-2,0]), leaving the top half white.

### DIFF
--- a/src/libdecaf/src/gpu/opengl/glsl2_alu.cpp
+++ b/src/libdecaf/src/gpu/opengl/glsl2_alu.cpp
@@ -467,13 +467,6 @@ insertDestBegin(fmt::MemoryWriter &out,
       flags = getInstructionFlags(inst.op3.ALU_INST());
    }
 
-   if (writeMask) {
-      auto gpr = inst.word1.DST_GPR().get();
-      out << "R[" << gpr << "].";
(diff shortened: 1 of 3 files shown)


brett19:
👍 lgtm, let's see what @exjam says.

exjam:
Makes sense, would be cool if it fixed some bugs!

Merged f67d9c85afc288b6180ca1afd13724cea78fe54b

PR #178, opened 2016-07-14
opengl: Handle CUBE instruction with non-xyz operands.

### LINKED ISSUE #177: CUBE with incorrect swizzling
From Mario Kart 8.

http://hastebin.com/raw/naxodamuji

@achurch can you take a look please

achurch:
Fixes #177 (I hope).


achurch:
Actually, I just noticed another instruction in that sequence that uses PS (whaaat). Need to fix that as well.

achurch:
Okay, this should take care of everything in the reported shader. I've also allowed ABS and NEG since if using PS for one of the operands works, presumably any sort of input is fine, and I can imagine at least NEG might make sense in some contexts.

PR #180, opened 2016-07-14
debugger: Make default visibility of windows configurable.

achurch:
I feel like this is something ImGui ought to save and restore along with window position/size, but based on the existing code it looks like that's not possible?

### DIFF
--- a/src/decaf-cli/config.cpp
+++ b/src/decaf-cli/config.cpp
@@ -31,7 +31,14 @@ struct CerealDebugger
    {
       using namespace decaf::config::debugger;
       ar(CEREAL_NVP(enabled),
-         CEREAL_NVP(break_on_entry));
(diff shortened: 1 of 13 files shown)


brett19:
I believe there is a way to handle this through the ImGui config storage and I think we should use that.  Note that I am attempting to avoid putting debugger::ui specific things in the config.  It's meant to be entirely self contained.

P.S. I suggest we close this PR for now.

achurch:
SGTM -- if ImGui has a way to do it then that's definitely cleaner.

exjam:
Looks like imgui only supports Pos, Size, Collapsed. So we'd have to fork imgui to add more config options if we want to use the imgui config storage...

https://github.com/ocornut/imgui/blob/master/imgui.cpp#L2380-L2424

PR #181, opened 2016-07-14
debugger: Properly select the interrupting thread on pause.

achurch:
The debugger still doesn't select the thread with the breakpoint/exception correctly because it never records which core initiated the pause (it looks like there was a variable intended for that purpose, but it was never used anywhere). This should get things working in all cases.


brett19:
Hey @achurch,
We've actually been intentionally avoiding any dependancy on ::debugger::ui from ::debugger.  I also noticed that our use of sActiveCore is a little messed up.  I'll fix this all in a comment in a moment.
Cheers, Brett

brett19:
Fixed through d9623a27.

achurch:
LGTM, thanks.

PR #182, opened 2016-07-15
decaf-sdl: Fix emulated screens not fitting inside the window.

achurch:
The TV and DRC screens each overflowed vertically by 18 pixels.


brett19:
👍 lgtm

PR #183, opened 2016-07-15
decaf-sdl: Antialias scan buffers when drawing them to the window.

achurch:
I don't know if the current behavior is intentional, but antialiasing makes it easier to see details (like small text) on the emulated screens given the default window size.


brett19:
Can you show us a screenshot comparison of the difference?

achurch:
Aliased:
Aliased

Antialiased:
Antialiased

brett19:
Could you move the sampler bind with the rest of the draw binding (where it originally was binding 0).  I know it makes more sense to do this once and draw many times, but we need a separate commit to move them.

achurch:
Sure thing, done.

PR #184, opened 2016-07-15
opengl: Use separate framebuffer objects for clear operations.

achurch:
I noticed this in the process of tracking down why Xenoblade only renders some primitives to the left 768 pixels of the TV screen. This patch doesn't solve that original problem, but it does fix some cases of depth buffers not being fully cleared because a smaller color buffer happened to be attached to GL_FRAMEBUFFER at the same time.


MoochMcGee:
@achurch YOU'RE SENDING SO MANY PULL REQUESTS! CALM DOWN

jk

brett19:
👍 lgtm, any better ideas @exjam?

PR #186, opened 2016-07-15
opengl: Return correct data type for R11G11B10 formats.

achurch:


exjam:
Are we meant to use gl::GL_UNSIGNED_INT_10F_11F_11F_REV for one of the orders and gl::GL_R11F_G11F_B10F for the other?

Which is the type you are seeing which this fixes? FMT_10_11_11_FLOAT or FMT_11_11_10_FLOAT?

achurch:
It's not an issue of bit order, it's that R11F_G11F_B10F is a format enum rather than a data type enum and isn't valid in the first place (it causes GL_INVALID_ENUM).

PR #189, opened 2016-07-16
coreinit: Enable locked cache DMA by default.

achurch:
Xenoblade never calls LCEnableDMA() (the rpx doesn't even reference it, for that matter) but uses LC{Load,Store}DMABlocks() for block copies.


brett19:
Why don't you join us on IRC?

brett19:
LC DMA is not enabled by default, its just not meant to be checked by the LC DMA calls.  Commit inbound.

brett19:
196c1a8b

PR #190, opened 2016-07-16
opengl: Fix disassembly of VTX_FETCH instructions in shader comments.

achurch:
The emulator still dies because it can't translate VTX_FETCH, but one step at a time...


brett19:
Lol.  Woops!

PR #192, opened 2016-07-16
Store configuration files in a platform-standard directory.

achurch:
Mainly so the system doesn't get littered with config.json and imgui.ini files in every directory you run the emulator from. I'll need someone else to write and test the code for Windows since I don't have a Windows build environment.

### DIFF
--- a/src/common/src/platform_posix_dir.cpp
+++ b/src/common/src/platform_posix_dir.cpp
@@ -1,7 +1,10 @@
 #include "platform.h"
 #include "platform_dir.h"
+#include "strutils.h"
 
 #ifdef PLATFORM_POSIX
+#include <errno.h>
+#include <stdlib.h>
 #include <sys/stat.h>
 #include <unistd.h>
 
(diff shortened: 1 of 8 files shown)


src/common/src/platform_posix_dir.cpp
```diff
@@ -46,6 +65,24 @@ isDirectory(const std::string &path)
    return S_ISDIR(info.st_mode);
 }
 
+std::string
+makeConfigPath(const std::string &filename)
```
brett19: The common project is meant to contain things which are not specific to decaf at all.  You should move this up to the emulator instead, and simply provide a method to retrieve the users HOME.

brett19:
I am alright with the idea of this commit, needs some minor fixing.  Let's see what @exjam has to say.

src/common/src/platform_posix_dir.cpp
achurch: Done, PTAL.

exjam:
Seems like a pretty good change, although might be confusing now as we do not haev a GUI to edit the config.

I committed with a few changes:
Windows implementation: e0076a55419e7c40f6517a70d821a663015ad1f3
Use HostPath: cd7ed3a6f8d9ab59b9d2463fa22cbb1ccbf0e95f

And instead of adding another group "Configuration Options" for the --config arg i just moved it to the Settings group. Also I made it so we log the config path (and game path): ad7e212c0cfe6f23648a264991f7ef14d64ca91e

PR #193, opened 2016-07-16
jit: Adjust call sequences to work on Linux.

achurch:
Linux uses the System V ABI, which differs from Windows in a few respects:
 - Register arguments are passed in RDI, RSI, RDX, RCX, R8, R9
 - RDI and RSI are volatile (so I moved mem::base() from RSI to RBP)
 - All XMM registers are volatile
 - The 128-byte region immediately below SP (red zone) is volatile
 - No argument shadow space is required for subroutine calls


achurch:
Resolved by 47cfb5a through 636deed.

PR #195, opened 2016-07-18
Add a --time-scale option to control the emulated clock speed.

achurch:
Just as a stopgap until there's a way to sync the emulated clock to the emulated CPU.


exjam:
Yeah at some point we need to make the time in coreinit be the same than the timebase registers in libcpu.

PR #198, opened 2016-07-18
jit: Tweak prologue/epilogue.

achurch:
Please doublecheck the stack allocation reduction from 0x38 to 0x28 -- all the documentation I've seen says the shadow space size is 0x20 (4 64-bit registers) and I don't see any stack references in generated code, but let me know if this breaks anything on Windows.

### DIFF
--- a/src/common/src/platform_posix_exception.cpp
+++ b/src/common/src/platform_posix_exception.cpp
@@ -17,29 +17,53 @@ namespace platform
 {
 
 static std::vector<ExceptionHandler>
-gExceptionHandlers;
+sExceptionHandlers;
 
 static struct sigaction
-gOldSegvHandler;
+sSegvHandler;
 
 static struct sigaction
-gSegvHandler;
+sSystemSegvHandler;
+
+static struct sigaction
+sIllHandler;
+
+static struct sigaction
+sSystemIllHandler;
 
 static void
-segvHandler(int unused_signum, siginfo_t *info, void *context)
+dispatchException(Exception *exception,
+                  void *context,
+                  int signum,
+                  const struct sigaction *ourHandler,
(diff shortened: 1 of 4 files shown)


brett19:
I agree that reducing the stack allocation is a good idea, however it appears this commit does other things which are not so clear.  We already have the nop additions wrapped in JIT_DEBUG, which is useful in general to dilineate JIT sections, there is no reason to ifdef that based on platform.  Additionally, putting an ifdef around EDI/ESI will cause loss of stack alignment.

achurch:
>We already have the nop additions wrapped in JIT_DEBUG, which is useful in general to dilineate JIT sections, there is no reason to ifdef that based on platform.

Sure there is - it's annoying having to step through all those NOPs when debugging. (:

>Additionally, putting an ifdef around EDI/ESI will cause loss of stack alignment.

No it won't -- remember it's RDI/RSI and not EDI/ESI, so 2*8 = 16 which maintains alignment.

brett19:
Fair enough, perhaps we can add a second define which controls the insertion of the NOP's.  Good call on the RDI/RSI...  Apparently my brain is stuck a decade ago.

achurch:
Actually, I guess it would make sense to #ifdef based on _MSC_VER since that's specific to Visual Studio? That would also make the #ifdef match the comment immediately following it. I've done that and also added an explicit constant (so the #ifdef is in only one place); let me know what you think.

brett19:
I think some stuff out of this still needs to be merged, but it conflicted with the JIT verify PR, so I skipped it for now.  If you rebase it, I'll be happy to merge it, but I think most of it is already covered in the JIT verify commit.

achurch:
Yeah, the JIT verifier covered most of it. Rebased the rest.

PR #200, opened 2016-07-19
Fix JIT bugs.

achurch:
The --jit-debug option didn't actually do anything, so I wrote a JIT verifier based on what the option said it should do. That turned up a number of JIT bugs, so I've fixed those.

### DIFF
--- a/src/libcpu/src/jit/jit.cpp
+++ b/src/libcpu/src/jit/jit.cpp
@@ -1,12 +1,15 @@
 #include "common/log.h"
+#include "common/align.h"
 #include "common/bitutils.h"
 #include "common/decaf_assert.h"
 #include "common/fastregionmap.h"
+#include "cpu.h"
 #include "cpu_internal.h"
 #include "espresso/espresso_instructionset.h"
 #include "jit.h"
 #include "jit_internal.h"
 #include "jit_insreg.h"
+#include "jit_verify.h"
 #include "jit_vmemruntime.h"
 #include "mem.h"
 #include <cfenv>
@@ -34,6 +37,12 @@ sJitBlocks;
 
 static uint8_t
 sBaseRelocCode[16];
+
+static void *
+gVerifyPreWrapper;
+
+static void *
+gVerifyPostWrapper;
 
 JitCall
 gCallFn;
@@ -65,19 +74,39 @@ initStubs()
    auto introLabel = a.newLabel();
    auto extroLabel = a.newLabel();
    auto exitLabel = a.newLabel();
+   auto verifyPreLabel = a.newLabel();
+   auto verifyPostLabel = a.newLabel();
 
    // This is invoked to set up the neccessary state for
    //  the JIT block we want to execute.
    a.bind(introLabel);
    a.push(asmjit::x86::rbp);
    a.push(asmjit::x86::rbx);
-   a.push(asmjit::x86::rdi);
+#ifdef PLATFORM_WINDOWS
+   a.push(asmjit::x86::rdi);  // RSI/RDI are callee-saved in Windows x64 ABI
    a.push(asmjit::x86::rsi);
+#endif
    a.push(asmjit::x86::r12);
    a.push(asmjit::x86::r13);
    a.push(asmjit::x86::r14);
    a.push(asmjit::x86::r15);
-   a.sub(asmjit::x86::rsp, 0x38);
+   if (gJitMode == jit_mode::verify) {
+      // Leave space for the verification buffer on the stack:
+      //  RSP+0 = Windows argument shadow space (reserved on all platforms for simplicity)
+      //  RSP+32 = size of this stack area (for reference on return)
+      //  RSP+36 = PPC instruction being verified
+      //  RSP+40 = verification buffer
+      uint32_t stackSpace = 40 + static_cast<uint32_t>(sizeof(VerifyBuffer));
+      // Preserve 16-byte stack alignment -- note that RSP comes in unaligned from the CALL instruction
+      stackSpace = align_up(stackSpace, 8);
+      if ((stackSpace & 15) == 0) {
+         stackSpace += 8;
+      }
+      a.sub(asmjit::x86::rsp, stackSpace);
+      a.mov(asmjit::X86Mem(asmjit::x86::rsp, 32, 4), stackSpace);
+   } else {
+      a.sub(asmjit::x86::rsp, 0x28);  // Windows shadow space + stack alignment
+   }
    a.mov(a.stateReg, a.sysArgReg[0]);
    a.mov(a.membaseReg, static_cast<uint64_t>(mem::base()));
    a.jmp(a.sysArgReg[1]);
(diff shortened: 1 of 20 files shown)


src/libcpu/src/jit/jit.cpp
```diff
@@ -35,6 +38,12 @@ sJitBlocks;
 static uint8_t
 sBaseRelocCode[16];
 
+static void *
+gVerifyPreWrapper;
```
brett19: Why don't we make this more generic as "gPreInstr, gPostInstr"

src/libcpu/src/jit/jit.cpp
```diff
@@ -65,19 +74,39 @@ initStubs()
...
+   auto verifyPreLabel = a.newLabel();
+   auto verifyPostLabel = a.newLabel();
 
    // This is invoked to set up the neccessary state for
    //  the JIT block we want to execute.
    a.bind(introLabel);
    a.push(asmjit::x86::rbp);
    a.push(asmjit::x86::rbx);
-   a.push(asmjit::x86::rdi);
+#ifdef PLATFORM_WINDOWS
```
brett19: We should not need to depend on an ifdef for save/load here.  Just use the a.stateReg, a.membaseReg variables, along with the pool listing available at like... a.regs or something.

src/libcpu/src/jit/jit.cpp
```diff
@@ -65,19 +74,39 @@ initStubs()
...
+      //  RSP+0 = Windows argument shadow space (reserved on all platforms for simplicity)
+      //  RSP+32 = size of this stack area (for reference on return)
+      //  RSP+36 = PPC instruction being verified
+      //  RSP+40 = verification buffer
+      uint32_t stackSpace = 40 + static_cast<uint32_t>(sizeof(VerifyBuffer));
+      // Preserve 16-byte stack alignment -- note that RSP comes in unaligned from the CALL instruction
+      stackSpace = align_up(stackSpace, 8);
+      if ((stackSpace & 15) == 0) {
+         stackSpace += 8;
+      }
```
brett19: What is this?  `stackSpace = align_up(stackSpace, 16) + 8` please.

src/libcpu/src/jit/jit.cpp
```diff
@@ -65,19 +74,39 @@ initStubs()
...
+      a.sub(asmjit::x86::rsp, stackSpace);
```
brett19: This is very dangerous, the stack unwind meta-data generator relies on constants for this stuff.  We might need to make it so the stack size selected is passed in or something.

src/libcpu/src/jit/jit.cpp
```diff
@@ -99,20 +128,105 @@ initStubs()
    a.bind(exitLabel);
    a.mov(a.niaMem, a.finaleNiaArgReg);
    a.mov(asmjit::x86::rax, a.stateReg);
-   a.add(asmjit::x86::rsp, 0x38);
+   if (gJitMode == jit_mode::verify) {
+      a.mov(asmjit::x86::ecx, asmjit::X86Mem(asmjit::x86::rsp, 32, 4));
```
brett19: this should be a constant value in verify mode even no?

src/libcpu/src/jit/jit.cpp
```diff
@@ -99,20 +128,105 @@ initStubs()
...
+      a.add(asmjit::x86::rsp, asmjit::x86::rcx);
+   } else {
+      a.add(asmjit::x86::rsp, 0x28);
+   }
    a.pop(asmjit::x86::r15);
    a.pop(asmjit::x86::r14);
    a.pop(asmjit::x86::r13);
    a.pop(asmjit::x86::r12);
+#ifdef PLATFORM_WINDOWS
```
brett19: As above, these save/loads should not need ifdefs

src/libcpu/src/jit/jit.cpp
achurch: That's going to add an extra 16 bytes if stackSpace = 0x...8.

src/libcpu/src/jit/jit.cpp
```diff
@@ -265,16 +379,25 @@ gen(JitBlock &block)
          a.bind(targetIter->second.label);
       }
 
-      if (JIT_DEBUG) {
```
brett19: This is needed for the debugger to function properly when access violations occur.

src/libcpu/src/jit/jit.cpp
```diff
@@ -265,16 +379,25 @@ gen(JitBlock &block)
...
-      }
-
       auto instr = mem::read<espresso::Instruction>(lclCia);
       auto data = espresso::decodeInstruction(instr);
       auto genSuccess = false;
 
       a.genCia = lclCia;
 
+      // Don't attempt to verify non-repeatable instructions
+      bool canVerify = (data->id != espresso::InstructionID::kc
```
brett19: Probably should move this inside the if statement below to avoid the cost of calculating it when the verifier is off.

src/libcpu/src/jit/jit.cpp
achurch: Yes -- this was just more convenient since there were 4 unused bytes anyway (due to alignment stuff).

src/libcpu/src/jit/jit.cpp
achurch: See comment on next line.

src/libcpu/src/jit/jit.cpp
brett19: Probably should just re-order everything around the call to fptr, such that its like...
```c++
if (JIT_DEBUG) {
  // save nia
}
auto fptr = ...;

if (!verify_mode) {
  fptr();
} else {
  insertPreVerify();
  fptr();
  insertPostVerify();
}
```

src/libcpu/src/jit/jit.cpp
achurch: Will do. Also g -> s since they're static (I originally had them extern but changed to static later).

src/libcpu/src/jit/jit_internal.h
```diff
@@ -108,6 +110,7 @@ class PPCEmuAssembler : public asmjit::X86Assembler
       PPCMemRef(lrMem, lr);
       PPCMemRef(ctrMem, ctr);
 
+      PPCMemRef(ciaMem, cia);
```
brett19: cia is intentionally not meant to be available through the assembler as it is meant only for debugging purposes and someday should be removed.  Use a.genCia (constant) instead.

src/libcpu/src/jit/jit_internal.h
```diff
@@ -193,6 +196,7 @@ class PPCEmuAssembler : public asmjit::X86Assembler
    asmjit::X86Mem lrMem;
    asmjit::X86Mem ctrMem;
 
+   asmjit::X86Mem ciaMem;
```
brett19: See above regarding use of CIA.

src/libcpu/src/jit/jit.cpp
achurch: It's still there, just moved.

src/libcpu/src/jit/jit_verify.h
```diff
@@ -0,0 +1,38 @@
...
+namespace cpu
+{
+
+namespace jit
+{
+
+struct
+VerifyBuffer
+{
+   Core coreCopy;               // Copy of core state before JIT execution
```
brett19: Core should never be copied.  Copy CoreRegs instead.

src/libcpu/src/jit/jit.cpp
achurch: I'd presume the compiler can optimize out the tests, but I can rearrange them.

src/libcpu/src/jit/jit_loadstore.cpp
```diff
@@ -85,6 +85,7 @@ loadGeneric(PPCEmuAssembler& a, Instruction instr)
          auto dst = a.loadXmmRegisterReadWrite(a.fprps[instr.rD]);
          a.movq(tmp, data);
          a.cvtss2sd(dst, tmp);
+         a.movddup(dst, dst);  // Copy to ps1 as well
```
brett19: The documentation explicitly indicates that it is not supposed to update ps1.

src/libcpu/src/jit/jit.cpp
achurch: I should mention it's important that core->nia remain unchanged until after the pre-verify call, so it's not already updated for the Core instance the interpreter sees (which could potentially mask bugs).

src/libcpu/src/jit/jit_internal.h
achurch: Can't do that since it's needed at runtime, not translation time. I can pass it as an extra argument to the verify functions, but it seems to me that your description of cia as a debugging tool exactly matches the purpose to which it's being put here.

src/libcpu/src/jit/jit_loadstore.cpp
```diff
@@ -88,8 +88,13 @@ loadGeneric(PPCEmuAssembler& a, Instruction instr)
          a.movddup(dst, dst);  // Copy to ps1 as well
       } else {
          decaf_check(sizeof(Type) == 8);
-         auto dst = a.loadXmmRegisterReadWrite(a.fprps[instr.rD]);
```
brett19: load____ReadWrite indicates that the register will be both read and written.  You should be still using ReadWrite, even if you copy out to a temporary before you use it, though I think there is another solution we should be using...

brett19: Here is the better fix:
```c++
auto dst = a.loadXmmRegisterReadWrite(a.fprps[instr.rD]);
a.movlpd(dst, data);
```

src/libcpu/src/jit/jit_loadstore.cpp
achurch: Nope, lfs does update ps1. My tests verify this.

src/libcpu/src/jit/jit_integer.cpp
```diff
@@ -601,15 +601,24 @@ neg(PPCEmuAssembler& a, Instruction instr)
 {
    auto eaxLockout = a.lockRegister(asmjit::x86::rax);
 
-   auto dst = a.loadRegisterWrite(a.gpr[instr.rD]);
```
brett19: Please follow the existing pattern of using allocGpTmp(src) to avoid collisions rather than writing out all the if statements.

src/libcpu/src/jit/jit_float.cpp
```diff
@@ -340,75 +340,56 @@ frsp(PPCEmuAssembler& a, Instruction instr)
    return true;
 }
 
-template <bool ShouldNegate>
+template <bool ShouldAbs, bool ShouldNegate>
 static bool
-fabsGeneric(PPCEmuAssembler& a, Instruction instr)
+fmrGeneric(PPCEmuAssembler& a, Instruction instr)
```
brett19: Could you point out which of these registers is the one which is actually corrupting ps1?  They should all operate soely on the low double of the XMM.

src/libcpu/src/jit/jit_float.cpp
```diff
@@ -56,155 +56,112 @@ absXmmSd(PPCEmuAssembler& a, const PPCEmuAssembler::XmmRegister& reg)
    a.pand(reg, maskXmm);
 }
 
-template <bool ShouldTruncate>
```
brett19: As before, please indicate which Sd instruction is corrupting the high-order double of the Xmm.

src/libcpu/src/jit/jit_float.cpp
```diff
@@ -37,14 +37,14 @@ hostHasFMA3()
 }
 
 static void
-truncateToSingleSd(PPCEmuAssembler& a, const PPCEmuAssembler::XmmRegister& reg)
```
brett19: Commit looks good, but what is the difference between those terms?  In reality, the operations perform the rounding by truncating as far as I could tell (P.S. names were copied from Dolphin's JIT).

src/libcpu/src/jit/jit_pairedsingle.cpp
```diff
@@ -23,23 +49,30 @@ mergeGeneric(PPCEmuAssembler& a, Instruction instr)
...
-      shflFlags |= (1 << 1);
+   auto tmpSrcA = a.allocXmmTmp();
+   auto tmpSrcB = a.allocXmmTmp();
+   {
+      auto srcA = a.loadRegisterRead(a.fprps[instr.frA]);
+      auto srcB = a.loadRegisterRead(a.fprps[instr.frB]);
+      if (flags & MergeValue0) {
+         a.movapd(tmpSrcA, srcA);
+         a.shufpd(tmpSrcA, tmpSrcA, 1);
+         roundToSingleSd(a, tmpSrcA, tmpSrcA);
```
brett19: I do not believe that the interpreter truncates or rounds the values, why are we doing it here.

src/libcpu/src/jit/jit_integer.cpp
achurch: Done.

src/libcpu/src/jit/jit_fallback.cpp
```diff
@@ -33,8 +33,8 @@ bool jit_fallback(PPCEmuAssembler& a, espresso::Instruction instr)
       a.lock().inc(asmjit::X86Mem(asmjit::x86::rax, 0));
    }
 
-   a.mov(asmjit::x86::rcx, a.stateReg);
-   a.mov(asmjit::x86::rdx, (uint32_t)instr);
+   a.mov(a.sysArgReg[0], a.stateReg);
```
brett19: I knew I'd miss one somewhere! :D

src/libcpu/src/jit/jit_loadstore.cpp
achurch: Good call, done.

src/libcpu/src/jit/jit_float.cpp
achurch: "round" = round according to FPSCR rounding mode, "truncate" = discard low bits (i.e. round toward zero). The behavior is actually different for ps0 and ps1 -- see my tests.

src/libcpu/src/jit/jit_float.cpp
achurch: movddup explicitly writes both double slots.

src/libcpu/src/jit/jit_float.cpp
achurch: Again, movddup.

src/libcpu/src/jit/jit_pairedsingle.cpp
achurch: Yes it does, see https://github.com/decaf-emu/decaf-emu/blob/master/src/libcpu/src/interpreter/interpreter_pairedsingle.cpp#L461-L481

You may recall I wrote that extensive test suite, so I really do know what I'm doing. (:

src/libcpu/src/jit/jit.cpp
brett19: align_down(stackSpace, 16) + 24?

src/libcpu/src/jit/jit.cpp
brett19: It just makes people wonder why it was done this way :)

achurch:
Will split this out into separate PRs per Brett's request.

src/libcpu/src/jit/jit.cpp
brett19: I saw it after I scrolled down more.  Cheers.

src/libcpu/src/jit/jit.cpp
brett19: I agree, I just like to be succinct in the code as it aids understanding.

src/libcpu/src/jit/jit_verify.h
achurch: Thing is, we need a full Core to pass to the interpreter instructions.

src/libcpu/src/jit/jit.cpp
achurch: Why are you making this harder than it needs to be? (: I think the existing code is clear as to the intent even if it does take more than one line, while I actually can't tell (without a bit of thought) whether this suggestion works correctly.

brett19:
```C++
if ((stackSpace & 15) == 0) {
   stackSpace += 8;
}
```
The intention of this bit is just a bit obfuscated (aligned to 16, + 8).  Just trying to find some way to make it more clear.

src/libcpu/src/jit/jit_verify.h
brett19: Hmm.  I think the best option would be to just create a Core object at the interpreter call site and drop the CoreRegs into it, but I guess having Core could work as long as you only copy the CoreReg's part.  The point of this is that Core is meant to be non-copyable (but havn't gotten there yet).

PR #201, opened 2016-07-19
jit: Implement JIT verifier.

achurch:


brett19:
Can you rebase this onto master, and try to move as much stuff out of jit.cpp into something like jit_verify.cpp?  I actually really need this right now, would be awesome to get it merged.

achurch:
Rebased. I've already written it so that most of the verification-specific logic is in jit_verify.cpp; what's left in jit.cpp is the code to interface to that.

brett19:
`/jit/jit.cpp` gets an extra 150+ lines of code just to interface?  That's saddening.  Someday we will need to come up with some way to do cool JIT things without writing code throughout.

brett19:
Is the tweaked prologue/epilogue supposed to go before or after this commit.  Or is it already closed?

brett19:
Your rebase doesn't actually build haha.  Trying to fix it now.

brett19:
Merged as bc1e5cb5

PR #204, opened 2016-07-19
jit: Fix lfd clearing the ps1 slot.

achurch:


brett19:
There is a better fix for this somewhere I wrote.  I'll find it and merge that instead.  Leaving this open for now for tracking purposes.

brett19:
b2dbcf6c47bc338f663a3a950a11eeaf75effa4b

PR #205, opened 2016-07-19
jit: Fix neg rN,rN destroying the value of rN.

achurch:


brett19:
I don't understand why this destroys the value of rN.

achurch:
I didn't trace through the code, but I suspect the problem is that loadRegisterWrite() overwrites the association between guest register rN and the host GPR to the newly allocated write-only register, so the following neg instruction negates that (undefined) value of that register instead of the register which held the original value of rN.

brett19:
It's wierd since in the case that rA and rD are the same, the generated code should look something like this (keep in mind that loadRegister always returns the same host register for any active PPC register, read/write only decides whether it needs to load the data from memory):
```asm
MOV EAX, EAX
NEG EAX
```

brett19:
Note also that loading a register for write, and then later read, it will read it first as long as the read usage proceeds the write usage.  Though in hindsight, its possible the precise ordering in this method breaks that logic.

brett19:
Yea, actually, looking at how this works, thats the problem.  We need to put the LOAD earlier in the function, because the way its written now, dst is used before the load because of function parameter calling order.

brett19:
Note to self: Might be good to make double-use of a register like this more explicitly written, and if there is a conflict to warn about it.

brett19:
d78d2bead0967cb6c0937c04dcd131d1a7dd41ac

PR #206, opened 2016-07-19
jit: Fix fmr/fneg/fabs/fnabs writing to ps1.

achurch:


brett19:
Espresso FPU cannot make up its mind on whether to touch PS1...

achurch:
Yeah, the 750CL paired single stuff is weird. I don't blame IBM for throwing it away in favor of real SIMD registers -- I just wish they'd gotten around to that before Nintendo picked that particular architecture...

brett19:
51af24efe3aebfc07349aa010ca7911436d8d10c

PR #207, opened 2016-07-19
jit: Fix double-precision arithmetic insns improperly writing to ps1.

achurch:


brett19:
Seems reasonable, I believe this means our JIT code will more closely match Interp.

P.S. So arithmetic instructions only copy to PS1 when they round... Wierd!

brett19:
669f4eb4e9b6fad5223cf57d33f14880b0de01bb

PR #208, opened 2016-07-19
jit: Fix ps1 handling in fused multiply-add instructions.

achurch:
Also use Intel FMA3 instructions when available to avoid loss of precision in the result.


brett19:
Nice!  This will be slightly faster as well.

brett19:
e69dd51a9abba8a771b62c1a99e1dc57c0b735a6

PR #209, opened 2016-07-19
jit: Fix misleading function names.

achurch:
"Truncate" and "round" are different operations, which is important for paired-single operations.

Note that this will conflict with the other changes in jit_float.cpp, so be careful when merging.


brett19:
Fair commit, will need to fix some other stuff related to this (ShouldTruncate -> ShouldRound).

brett19:
e0aa42dd

PR #210, opened 2016-07-19
jit: Fix rounding of values in ps_merge.

achurch:


brett19:
I think there is a better way to do this rather than `shufpd`'ing everything!

achurch:
For the first slot, we could potentially use cvt*pd (instead of sd) as long as a SNaN in the low double doesn't abort the operation on the high double (which I haven't checked). You seemed to prefer avoiding special cases, though, so I wrote the code to pass things through the existing rounding function.

For the second, I don't think there's any way that would take fewer instructions -- you have to either move the high double down to the low end or move the mask up to the high end. (Or you load the 16-byte mask from memory somewhere, but that adds an unnecessary memory access.)

brett19:
It seems to me that you should be able to use only a single temporary for whatever ends up in the high double, and the lower double can be truncated/rounded in-place.

P.S. For future reference, I prefer avoiding code duplication and special cases are okay depending on the context, if its just in a single function for the purpose of improving how it works, thats probably alright.  I'm just against having special casing be used in every single JIT function, at that point it should just be solved generically at a higher level.

brett19:
328802c7d8c69b7884cc6a05713b8767f2463ce8

PR #212, opened 2016-07-19
interpreter: Fix inaccuracies in ps_res handling.

achurch:
Use ppc_estimate_reciprocal() instead of regular division, and avoid setting FPSCR[XX] on inexact results.

### DIFF
--- a/src/libcpu/src/interpreter/interpreter_pairedsingle.cpp
+++ b/src/libcpu/src/interpreter/interpreter_pairedsingle.cpp
@@ -538,7 +538,7 @@ ps_res(cpu::Core *state, Instruction instr)
       } else if (vxsnan0) {
          d0 = make_nan<float>();
       } else {
-         d0 = 1.0f / static_cast<float>(b0);
+         d0 = ppc_estimate_reciprocal(static_cast<double>(static_cast<float>(b0)));
       }
       updateFPRF(state, d0);
    }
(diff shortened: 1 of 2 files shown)


src/libcpu/src/interpreter/interpreter_pairedsingle.cpp
```diff
@@ -559,7 +559,12 @@ ps_res(cpu::Core *state, Instruction instr)
       state->fpr[instr.frD].paired1 = extend_float(d1);
    }
 
+   // ps_res never sets FPSCR[XX], so avoid setting it here.
+   const int xx = state->fpscr.xx;
    updateFPSCR(state, oldFPSCR);
+   if (!xx) {
```
brett19: According to the documentation, XX is not cleared, it is just untouched.  This should be:
```c++
state->fpsrc.xx = xx;
```

achurch: I actually said exactly that in the comment above. (: Since updateFPSCR() can only set XX, not clear it, the code is correct as written. But I agree your way is more concise (and possibly faster on average), so I'll update it when I'm back at the code, or feel free to make the change yourself when merging.

brett19: Ahh.  I didn't realize that updateFPSRC did an OR into XX.

brett19:
616f7b14fd2477a5fc3c5894a1f22310d339e6d6

PR #214, opened 2016-07-19
interpreter: Work around a stupid LLVM bug.

achurch:
At the moment there only seems to be one affected function, and the workaround turned out to be just a matter of disabling optimization on the function, so I went ahead and did that, as ugly as it looks.

### DIFF
--- a/src/common/platform_compiler.h
+++ b/src/common/platform_compiler.h
@@ -1,0 +1,13 @@
+#pragma once
+
+// Macro to disable optimization when building with Clang on a function which
+//  both performs floating-point operations and checks exception flags that
+//  could be affected by those operations.  This is required because LLVM is
+//  unaware of the fact that floating-point operations can raise exceptions
+//  (see http://llvm.org/bugs/show_bug.cgi?id=6050).  Place this immediately
+//  before the opening brace for the function.
+#ifdef __clang__
+#define CLANG_FPU_BUG_WORKAROUND __attribute__((optnone))
+#else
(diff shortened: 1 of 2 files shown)


brett19:
While I understand this commit, I think its a bad idea to assume that this function is the only place that this kind of optimization failure would occur.  It might be worth placing a common define in the `common/platform_x.h` files which you can use on any method to indicate that it does FPU exception stuff that needs to be deopt'd on clang.

achurch:
This isn't so much "assuming it's the only place it occurs" as "correcting an existing bug". In testing with Xenoblade, no other JIT instructions reported mismatches, and while I didn't actually track which instructions were hit, I did touch pretty much all the FP-related code, so it seems unlikely there are any similar bugs lurking at the moment.

That said, I agree it would be cleaner to have a #define so we don't need to clutter actual source code with this. I'll update this when I have a chance, or if you get to merging first, feel free to fix it yourself.

brett19:
Fair enough.  I just meant that this bug is completely at the whim of the optimizer.  So while it doesn't appear to be affecting anything else, at any point in time, it may.  We should find a more generic solution we can insert at every case where its possible the optimizer might break something.

achurch:
Updated, PTAL.

brett19:
Interpreter fres and updateFPSCR also use the std floating point exception functions.  These could break if the optimizer rearranges them around at all as well.  We should put the FPU fix there as well.  Otherwise we are just at the whim of the optimizer.

achurch:
Neither of those functions perform floating-point operations, so they're not affected by this bug.

PR #216, opened 2016-07-23
kernel: Skip normal interrupt processing for breakpoints.

achurch:
Breakpoints are not masked by OSDisableInterrupts(), so if a core was stopped with interrupts disabled, the cpu::this_core::interruptMask() == cpu::INTERRUPT_MASK assertion in coreinit::internal::checkRunningThreadNoLock() would fail when the core resumed.

For me, this triggered sporadically when trying to step in the debugger after hitting a breakpoint.

### DIFF
--- a/src/libdecaf/src/kernel/kernel.cpp
+++ b/src/libdecaf/src/kernel/kernel.cpp
@@ -207,7 +207,8 @@ cpuInterruptHandler(uint32_t interrupt_flags)
       coreinit::internal::pauseCoreTime(false);
    }
 
-   if (!(interrupt_flags & ~cpu::NONMASKABLE_INTERRUPTS)) {
+   auto unsafeInterrupts = cpu::NONMASKABLE_INTERRUPTS | cpu::DBGBREAK_INTERRUPT;
+   if (!(interrupt_flags & ~unsafeInterrupts)) {
       // Due to the fact that non-maskable interrupts are not able to be disabled
       // it is possible the application has the scheduler lock or something, so we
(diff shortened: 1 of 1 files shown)


brett19:
Not good.  This will drop interrupts if they come in together.

achurch:
Uh, I'm not touching the NONMASKABLE_INTERRUPTS bit.

brett19:
I mean to fix this if statement, which had the entire point of solving the issue you are mentioning.  Its just that we had to change what interrupts were maskable.

```c++
    if (!(interrupt_flags & ~cpu::NONMASKABLE_INTERRUPTS)) {
       // Due to the fact that non-maskable interrupts are not able to be disabled
       // it is possible the application has the scheduler lock or something, so we
       // need to stop processing here or else bad things could happen.
       return;
    }
```

achurch:
So you want me to just remove that entirely? Because the if block I modified (which isn't visible in the diff) checks for exactly DBGBREAK_INTERRUPT.

brett19:
```c++
    auto unsafeInterrupts = cpu::NONMASKABLE_INTERRUPTS | cpu::DBGBREAK_INTERRUPT;
    if (!(interrupt_flags & ~unsafeInterrupts)) {
       // Due to the fact that non-maskable interrupts are not able to be disabled
       // it is possible the application has the scheduler lock or something, so we
       // need to stop processing here or else bad things could happen.
       return;
    }
```

achurch:
Done, PTA.

PR #218, opened 2016-07-23
Implement basic sound output via SDL.

achurch:
This gives me approximately the correct output for Xenoblade menu sound effects (ADPCM) and title BGM (streamed ADX, passed to AXVoice as linear PCM). I have no idea how well it will work with any other game.

This is all very much in the experimental stages -- feel free to suggest changes.

### DIFF
--- a/build/decaf-sdl.vcxproj
+++ b/build/decaf-sdl.vcxproj
@@ -18,6 +18,7 @@
     <ClCompile Include="..\src\decaf-sdl\config.cpp" />
     <ClCompile Include="..\src\decaf-sdl\decafsdl_draw.cpp" />
     <ClCompile Include="..\src\decaf-sdl\decafsdl_input.cpp" />
+    <ClCompile Include="..\src\decaf-sdl\decafsdl_sound.cpp" />
     <ClCompile Include="..\src\decaf-sdl\main.cpp" />
     <ClCompile Include="..\src\decaf-sdl\decafsdl.cpp" />
   </ItemGroup>
@@ -25,6 +26,7 @@
     <ClInclude Include="..\src\decaf-sdl\config.h" />
     <ClInclude Include="..\src\decaf-sdl\decafsdl.h" />
     <ClInclude Include="..\src\decaf-sdl\clilog.h" />
+    <ClInclude Include="..\src\decaf-sdl\decafsdl_sound.h" />
   </ItemGroup>
   <PropertyGroup Label="Globals">
     <ProjectGuid>{E10BC5AA-C1A4-4B17-943D-9C1B4BCD52A9}</ProjectGuid>
--- a/build/decaf-sdl.vcxproj.filters
+++ b/build/decaf-sdl.vcxproj.filters
@@ -30,6 +30,9 @@
     <ClCompile Include="..\src\decaf-sdl\decafsdl_input.cpp">
       <Filter>Source Files</Filter>
     </ClCompile>
+    <ClCompile Include="..\src\decaf-sdl\decafsdl_sound.cpp">
+      <Filter>Source Files</Filter>
+    </ClCompile>
   </ItemGroup>
   <ItemGroup>
     <ClInclude Include="..\src\decaf-sdl\config.h">
@@ -39,6 +42,9 @@
       <Filter>Header Files</Filter>
     </ClInclude>
     <ClInclude Include="..\src\decaf-sdl\decafsdl.h">
+      <Filter>Header Files</Filter>
+    </ClInclude>
+    <ClInclude Include="..\src\decaf-sdl\decafsdl_sound.h">
       <Filter>Header Files</Filter>
     </ClInclude>
   </ItemGroup>
--- a/build/libdecaf.vcxproj
+++ b/build/libdecaf.vcxproj
@@ -157,6 +157,7 @@
     <ClCompile Include="..\src\libdecaf\decaf_nullgraphicsdriver.cpp" />
     <ClCompile Include="..\src\libdecaf\decaf_input.cpp" />
     <ClCompile Include="..\src\libdecaf\decaf_nullinputdriver.cpp" />
+    <ClCompile Include="..\src\libdecaf\decaf_sound.cpp" />
     <ClCompile Include="..\src\libdecaf\src\debugger\debugger.cpp" />
     <ClCompile Include="..\src\libdecaf\src\debugger\debugger_analysis.cpp" />
     <ClCompile Include="..\src\libdecaf\src\debugger\debugger_ui.cpp" />
@@ -385,6 +386,7 @@
     <ClInclude Include="..\src\libdecaf\decaf_input.h" />
     <ClInclude Include="..\src\libdecaf\decaf_nullgraphicsdriver.h" />
     <ClInclude Include="..\src\libdecaf\decaf_nullinputdriver.h" />
+    <ClInclude Include="..\src\libdecaf\decaf_sound.h" />
     <ClInclude Include="..\src\libdecaf\src\debugger\debugger.h" />
     <ClInclude Include="..\src\libdecaf\src\debugger\debugger_analysis.h" />
     <ClInclude Include="..\src\libdecaf\src\debugger\debugger_branchcalc.h" />
--- a/build/libdecaf.vcxproj.filters
+++ b/build/libdecaf.vcxproj.filters
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <ItemGroup>
     <Filter Include="Source Files">
@@ -849,6 +849,9 @@
     <ClCompile Include="..\src\libdecaf\src\modules\coreinit\coreinit_screen.cpp">
       <Filter>Source Files\modules\coreinit</Filter>
     </ClCompile>
+    <ClCompile Include="..\src\libdecaf\decaf_sound.cpp">
+      <Filter>Source Files</Filter>
+    </ClCompile>
   </ItemGroup>
   <ItemGroup>
     <ClInclude Include="..\src\libdecaf\src\modules\coreinit\coreinit.h">
@@ -1594,6 +1597,9 @@
     </ClInclude>
     <ClInclude Include="..\src\libdecaf\src\modules\coreinit\coreinit_screen_font.h">
       <Filter>Header Files\modules\coreinit</Filter>
+    </ClInclude>
+    <ClInclude Include="..\src\libdecaf\decaf_sound.h">
+      <Filter>Header Files</Filter>
     </ClInclude>
   </ItemGroup>
   <ItemGroup>
--- a/src/decaf-cli/config.cpp
+++ b/src/decaf-cli/config.cpp
@@ -61,6 +61,16 @@ struct CerealJit
    }
 };
 
+struct CerealSound
+{
+   template <class Archive>
+   void serialize(Archive &ar)
+   {
+      using namespace decaf::config::sound;
+      ar(CEREAL_NVP(dump_sounds));
+   }
+};
+
 struct CerealSystem
 {
    template <class Archive>
@@ -88,6 +98,7 @@ bool load(const std::string &path)
    try {
       input(cereal::make_nvp("jit", CerealJit {}),
             cereal::make_nvp("log", CerealLog {}),
+            cereal::make_nvp("sound", CerealSound {}),
             cereal::make_nvp("system", CerealSystem {}));
    } catch (std::exception e) {
       // Can't use gLog because it is NULL here.
@@ -104,6 +115,7 @@ void save(const std::string &path)
    cereal::JSONOutputArchive output(file);
    output(cereal::make_nvp("jit", CerealJit {}),
           cereal::make_nvp("log", CerealLog {}),
+          cereal::make_nvp("sound", CerealSound {}),
           cereal::make_nvp("system", CerealSystem {}));
 }
 
--- a/src/decaf-sdl/config.cpp
+++ b/src/decaf-sdl/config.cpp
@@ -121,6 +121,16 @@ struct CerealJit
    }
 };
 
+struct CerealSound
+{
+   template <class Archive>
+   void serialize(Archive &ar)
+   {
+      using namespace decaf::config::sound;
+      ar(CEREAL_NVP(dump_sounds));
+   }
+};
+
 struct CerealSystem
 {
    template <class Archive>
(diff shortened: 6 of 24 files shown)


TechnoCrunch:
Would it be possible to add the audio settings to the config files?
Mainly the `audiospec.channels = 2; audiospec.samples = 1024;` lines, so they are configurable per game? And so other things can be tested like the Wii U's surround sound output and possibly Dolby digital audio calls.
You would have to just do it in [decaf_config.h](https://github.com/decaf-emu/decaf-emu/blob/master/src/libdecaf/decaf_config.h) and create your own namespace audio and do the rest.

achurch:
>Would it be possible to add the audio settings to the config files?

That sort of aesthetic stuff is for later work. The goal of this change is to try and get sound working in the first place.

exjam:
So I've been thinking about this and how it ties into the audio callbacks. I think we need to establish a proper audio pipeline. I went ahead and posted full list of exports from sndcore2.rpl and snduser2.rpl to the wiki.

https://github.com/decaf-emu/decaf-emu/wiki/sndcore2.rpl
https://github.com/decaf-emu/decaf-emu/wiki/snduser2.rpl

If we grep them for the word Callback we can see some interesting information:

```
$ cat sndcore2.rpl.txt  | grep Callback
    0x02017414 AI2RegisterDMACallback
    0x020177B0 AIDRCRegisterDMACallback
    0x02016BDC AIRegisterDMACallback
    0x0200C598 AXDeregisterAppFrameCallback
    0x02001DC8 AXGetAuxACallback
    0x02001DE4 AXGetAuxBCallback
    0x02001E00 AXGetAuxCCallback
    0x02001F98 AXGetAuxCallback
    0x02001E1C AXGetAuxDRCCallback
    0x0200B9D4 AXGetDeviceFinalMixCallback
    0x0200BA44 AXGetPostFinalMixCallback
    0x0200C4A4 AXRegisterAppFrameCallback
    0x02001BF8 AXRegisterAuxACallback
    0x02001C6C AXRegisterAuxBCallback
    0x02001CE0 AXRegisterAuxCCallback
    0x02001E38 AXRegisterAuxCallback
    0x02001D54 AXRegisterAuxDRCCallback
    0x0200B6E8 AXRegisterCallback
    0x0200B734 AXRegisterDRCCallback
    0x0200B95C AXRegisterDeviceFinalMixCallback
    0x0200B924 AXRegisterExceedCallback
    0x0200B6A0 AXRegisterFrameCallback
    0x0200BA08 AXRegisterPostFinalMixCallback
```

So we're really looking at
- Exceed Callback - this is something to do with the DSP
- Frame Callback - we know this is every audio frame 3ms
- App Frame Callback - we are fairly confident this is the same thing as Frame Callback
- AUX Callbacks (A, B, C, DRC) - some channel shit? could be mixing related
- Device Final Mix Callback - so final mix once we have audio for an output device?
- Post Final Mix Callback - a callback after final mix? dunno why they would need this, I assume final mix means the data is allll ready for sound output

Then in snduser2 it pretty much looks like its all just special affect callbacks, these are probably applied to voices. There is also SEQ and SYN which seems to be a midi sequencer and synthesizer.

```
$ cat snduser2.rpl.txt | grep Callback
    0x02007ADC AXFX2ChorusCallback
    0x02008528 AXFX2CompressorCallback
    0x02008E74 AXFX2DelayCallback
    0x0200942C AXFX2FlangerCallback
    0x02009C08 AXFX2OverdriveCallback
    0x0200A66C AXFX2PitchshiftCallback
    0x0200B0E0 AXFX2ReverbCallback
    0x0200BA64 AXFX2ReverbI3dl2Callback
    0x02012670 AXFX2chChorusCallback
    0x0201371C AXFX2chMultiDelayCallback
    0x02013EC8 AXFX2chReverbCallback
    0x020127D8 AXFX4chChorusCallback
    0x02013848 AXFX4chMultiDelayCallback
    0x02014040 AXFX4chReverbCallback
    0x02014330 AXFX6ch6chReverbCallback
    0x02012508 AXFX6chChorusCallback
    0x020135F0 AXFX6chMultiDelayCallback
    0x020141B8 AXFX6chReverbCallback
    0x0200BBCC AXFXChorusCallback
    0x0200BD10 AXFXChorusCallbackDpl2
    0x0200C5A8 AXFXChorusExpCallback
    0x0200D1E4 AXFXChorusExpCallbackDpl2
    0x0201009C AXFXDelayCallback
    0x0201073C AXFXDelayCallbackDpl2
    0x020110B8 AXFXDelayExpCallback
    0x02011AD0 AXFXDelayExpCallbackDpl2
    0x02012308 AXFXMultiChChorusCallback
    0x020134CC AXFXMultiChDelayCallback
    0x02013D4C AXFXMultiChReverbCallback
    0x020144C8 AXFXReverbHiCallback
    0x020145E4 AXFXReverbHiCallbackDpl2
    0x0201514C AXFXReverbHiExpCallback
    0x020161FC AXFXReverbHiExpCallbackDpl2
    0x020168A8 AXFXReverbStdCallback
    0x020169BC AXFXReverbStdCallbackDpl2
    0x02017430 AXFXReverbStdExpCallback
    0x020182B4 AXFXReverbStdExpCallbackDpl2
    0x0200740C SEQRegisterControllerCallback
    0x02004B20 SYNSetInitCallback
    0x02004B84 SYNSetUpdateCallback
```

So yeah what we need to establish is the pipeline, from voices to output, so basically voice -> fx -> mix -> output, but in the correct detail (and order?).

Also before we merge this as the "initial" audio pipeline, I'd like to see if we can invert the control flow here. Currently you have SDL asking the emulator for more audio, however SDL can only do that callback when it runs out of samples and the sample size is a power of 2 number. However we need to consume the AXVoice shit every 3ms inline with the frame callback. Note this may be why audio is fucked on some games because we are doing 1024 samples which is ~30ms whereas games would only expect to have to supply like 3ms of audio in the voices at any one time.

For now we can ignore the complex stuff like AXFX and midi shit, but at least try to get our voice -> mix -> output with at least the frame callback and pushing 3ms of audio through to SDL instead of a SDL callback.

achurch:
>Also before we merge this as the "initial" audio pipeline, I'd like to see if we can invert the control flow here.

I was about to say we can't, because SDL audio always uses the pull style, but it looks like SDL 2.0.4 added a push method so I'll look into that (may be a few days since I've got some other stuff to take care of). By side effect, that would make it a lot easier to have a WAV backend too, since all it would have to do is dump the pushed data to disk.

exjam:
Yeah SDL_QueueAudio. Hmm I didn't think about that - a WAV backend would be pretty useful for debugging  (and testing) too, especially if we make it possible to swap out audio backends at runtime. :+1: 

Would be cool if we could use the Post Final Mix callback to dump the final mixed audio from hardware with some unit tests and we can see how close to perfect our voice, fx, mixing stuff is. Although that would depend on if that is what that function is actually for or not - haven't really looked into it. Also writing tests is tedious... :-1:

hrydgard:
Just adapt from push to pull style by adding a little queue in between, just like Dolphin and PPSSPP do. Most really fast hardware APIs like WASAPI on Windows or CoreAudio on Mac are pull-style anyway, and you usually can't control the hardware buffer size with very much precision anyway so you can't push directly whatever you do, so might as well get it over with.

achurch:
Yeah, I'm well aware of how audio output works at a low level, but as long as SDL provides a push interface (and as long as it's not too broken) there's no reason to reinvent that wheel.

brett19:
Hey Guys,

I started to look at this, and discussed a bit with @exjam.  I'm not able to code the results of the discussions and reversing due to some time constraints at the moment, so it seemed prudent to record it here.  This commit is an amazing start to getting audio, we just need to rearrange the order of processing a bit before we commit it to the mainline (after the rearrange to be closer to what we expect to need in the future, we are happy to merge it even in spite of it being incomplete or down-right broken for some things).

Our current implementation of having the AX callback thread invoked every 3ms is on-par with the real WiiU, but this commit completely obviates its behaviour.  Obviously we still need to meld together with SDL which presents some issues as well.  Here is the proposal for a 'semi-working, mergable commit':

We should be using int32_t for our 'internal' audio data, this is what FinalMixCallback appears to receive, so its logical that we store our internal data in a way that is easy to pass there, SDL supports this inherently.  We need to add a new structure to snd_core next to the sVoiceExtras called sVoiceGenData which is statically allocated data used for audio generation.  This should have an array of 144 samples (int32_t), which is the samples used for 3ms of data at 48000khz (some of those samples may not be used when sampling at a lower rate).

At the beginning of the AX callback we need to scan through all active playing voices and read samples enough to populate their buffers, applying relevant voice 'effects' as well, this includes: rate conversion (VoiceSrc) [note that I think this affects how many samples we should read to generate the 144 or so output samples], envelopes (VoiceVe), LPF (VoiceLpf), BIQUAD (VoiceBiquad).  Once the samples for each voice are ready, we then need to mix these onto a new set of structures representing each device and bus.  We then need to invoke the AUX Callbacks (RegisterAuxCallback) for each of the AUX buses for each device.  Then our frame callbacks should be invoked (as we do now).  At this point, I believe we simply need to mix all of the buses together into per-device/per-channel streams and invoke the final mix callback for each device (not sure about how it receives data at the moment).  Once the final mix callbacks are completed, we should PUSH the 3ms of samples for each device (and channel?) to DecafSDL.  Note that the data pushed to `hardware` (ie: decaf::AudioDriver) should always be 48,000khz, and the upsampling stage appears to be controllable (maybe AXSetDeviceUpsampleStage?).

On the DecafSDL side of things, we should have a running buffer that is related to the sizes of the SDL hardware buffer sample count.  This buffer should be approximately 2x the hardware sample count.  The SDL callback should look at this buffer, and wait until its completely full (writing silence until that point).  Once the buffer is full, it should flip a flag indicating that it is reading from the buffer, and then read off enough data to fill the hardware buffer.  If at any point our internal buffer becomes completely empty, we should flip off our flag and stop copying to hardware.  This should make our audio implementation relatively resilient to various artifacts caused by libcpu being slow.

Additionally, I believe that our DecafSDL audio should be always active and attempting to play audio from the time that the emulator starts up (it should always be copying silence to the hardware buffers if no data was sent from libdecaf).  This is relevant for avoiding having to have audio-specific stuff in the debugger and such.  You should also be able to have DecafSDL inherit from a decaf::SoundDriver directly since it is a push-only interface (it always will push 144 int32_t samples per device, per channel)

P.S. In case you did not discover it yet, Nintendo's DSPADPCM format is used across all their consoles and you can find its exact implementation on a few sites, here is one I came across: http://www.metroid2002.com/retromodding/wiki/DSP_(File_Format) .

Cheers, Brett

achurch:
Hey Brett -- SGTM, thanks for all the suggestions! My schedule is still a bit random but I'll try and get the code done sometime in the next day or so.

With respect to this, though:

>Note that the data pushed to hardware (ie: decaf::AudioDriver) should always be 48,000khz

Are you sure about that? Xenoblade passes 1.0 (0x10000 fixed-point) to SetVoiceSrc for 32kHz samples and 1.5 (0x18000) for 48kHz samples, so that makes me want to say the final mix is 32kHz. I haven't done any actual tests so I don't have any solid grounds for doubting you, but it would seem strange to me if the real console had two resampling stages (48->32 and then 32->48) so I just wanted to know why you suggest this.

Also, the ADPCM link you gave doesn't deal with those high-bit-set header bytes so I guess those are still a mystery? (: I did find the algorithm by looking in FFmpeg, but thanks anyway.

brett19:
Hey @achurch,

I suspect the reason that you are seeing Xenoblade downsample its input audio to 32kHz is that they have initialized the audio system at 32kHz (presumably many games use 32kHz internal sampling for performance reasons). At the final stage, the output would be upsampled to the 48kHz that the hardware expects.  The main reason I expect this layout is that if the output hardware did not support 48kHz, then they would have 0 reason to even support it.  This, in addition to the apparent behaviour of SetUpsampleStage indicates that 48kHz would be the output rate.

Can you elaborate on those `high-bit header bytes`?  It seems to me that the format I linked to fully describes all bits of all bytes?

Cheers, Brett

exjam:
Well bare in mind AXInit is from snd_core.rpl which has no option to pass in anything other than 32khz, sndcore2.rpl introduced AXInitWithParams whichs allows you to pass in 48khz as an option.

Xenoblade uses the legacy AXInit, so maybe they made the game back before they had the option for 48khz.

achurch:
>At the final stage, the output would be upsampled to the 48kHz that the hardware expects.

Is the hardware known to support only 48kHz, or could it support both 32 and 48? Seems odd that Nintendo would have made AXInit support only a non-native sampling rate.

>Can you elaborate on those high-bit header bytes? It seems to me that the format I linked to fully describes all bits of all bytes?

In the ADPCM dump @exjam gave me (NES Remix, I think?) there were some bytes interspersed between the 8-byte ADPCM blocks that had the high bit set, and dropping those bytes from the stream resulted in (what sounded like) correct output. If you attempt to decode them as block header bytes then you get the crazy static from the earlier code.

exjam:
If the upsample is only for output to device, surely we do not need to bother doing it because SDL can accept either sample rate. Or is the upsampled data actually accessible to the game?

brett19:
The upsampled data can be made available to the game depending on the upsample stage.  It would definitely be more clear if we use a common pattern for upsampling, even if it happens to be afterwards.

achurch:
I've incorporated most of the suggestions; PTAL. (With the push pipeline, Xenoblade title screen BGM no longer skips, so good call!)

A couple of things I didn't do:

- I've left the output at 32kHz, since adding upsample logic would have no benefit and a possible drawback: currently game code can't see the difference, so adding code to upsample to 48kHz won't improve compatibility in any fashion, and the code might have to be rewritten anyway once the actual behavior of the output pipeline is confirmed.

- I've left DecafSDLSound as a separate class since it's not trivial, and also since it could reasonably be replaced by another driver -- WAV, for example, or stuff like PortAudio if somebody feels so strongly about it that they want to write the code for it...

brett19:
Hey @achurch,
What is extras->offsetFrac for?
Cheers, Brett

achurch:
extras->offsetFrac counts partial samples for resampling. It increments by the value set with AXSetVoiceSrc (or AXSetVoiceSrcRatio * 0x10000) for each sample played, and each 0x10000 increment is equivalent to advancing by one sample. So setting 0x10000 plays the sample back at the output rate, 0x18000 plays back at 1.5x (for playing 48kHz samples on 32kHz output), etc.

achurch:
Oh, one thing I just realized I forgot to mention -- remember how we were talking about SDL_QueueAudio? Turns out it doesn't work. :P At least on Linux, I couldn't get the new audio device API (SDL_OpenAudioDevice etc., of which SDL_QueueAudio is a part) to give me any sound output whether I used the push or pull style, so I just went with the traditional SDL_OpenAudio interface instead.

brett19:
Having our own AX callback for how to handle reading from our buffer is probably the best option anyways.  Especially since we can handle the 'catchup silence' better that way.

PR #220, opened 2016-07-24
Mask cb_shader_mask to select only the bits for the current export

Rozelette:
Fixes a problem for a shader in Xenoblade. It exports to export 0, but only export 2 is enabled. Because the higher order bits of cb_shader_mask are not masked out, the fact that a higher numbered export is enabled causes the check if the current export is fully masked to fail.

Idk if this is being caused by another issue somewhere (several asserts have to be ignored to get Xenoblade to the point when this shader is compiled), but to me this seems like a bug regardless.


exjam:
Definitely is a bug. Although maybe we should remove the warning there.

PR #238, opened 2016-08-03
opengl: Implement GET_TEXTURE_INFO.

achurch:
Xenoblade uses this. I don't know how the instruction selects the mipmap to return info for; at the moment I use the first component of the input vector.

### DIFF
--- a/src/libdecaf/src/gpu/opengl/glsl2_tex.cpp
+++ b/src/libdecaf/src/gpu/opengl/glsl2_tex.cpp
@@ -9,7 +9,6 @@ VTX_FETCH
 VTX_SEMANTIC
 MEM
 LD
-GET_TEXTURE_INFO
 GET_SAMPLE_INFO
 GET_COMP_TEX_LOD
 GET_GRADIENTS_H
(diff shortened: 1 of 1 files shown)


Rozelette:
I think that texTmp should be padded out with some default value to always be a vec3. I don't have any examples of this in games, but to me it look like that the ISA allows getting the depth of, say, a 1D texture (even though GLSL doesn't). In that case, to me it would make sense to get a value of 1, but idk what the hw actually does. Maybe for now there should be an assert that numDstSels <= samplerElements ?

Also, for textureQueryLevels you need to enable the GL_ARB_texture_query_levels extension.

achurch:
>I think that texTmp should be padded out with some default value to always be a vec3.

Done. I went with 1 as you suggest, since that would make the most sense.

>Also, for textureQueryLevels you need to enable the GL_ARB_texture_query_levels extension.

I don't think so? The GLSL 4.20 spec includes it.

exjam:
There's no reason we had #version 420 other than dank memes :fire:

achurch:
Hah, fair enough (:
I've bumped it to 450 in the streamout change so all should be well.

PR #239, opened 2016-08-03
jit: Fix an array overrun

lioncash:
Based off initialization code, this is actually supposed to be 32 bytes in size.


degasus:
This PR fixes the JIT on linux.

Issue #234, opened 2016-08-02
Implement Stream Out / Transform Feedback shaders


achurch:
Xenoblade gives this assertion failure for the shader pair below. The vertex shader maps semantic IDs 64-66 to outputs 0-2, but the pixel shader looks up semantic ID 0, triggering the assertion.

I note that the vertex shader has a POS0 output in addition to PARAM0-PARAM2; could this be what the pixel shader is trying to read from? (Though on second thought, that doesn't make sense because the POS0 output is a constant {0,0,0,1}...)

Vertex shader: (this looks like it's actually the same shader as in issue #221)
```
00 CALL_FS NO_BARRIER

01 ALU ADDR(32) CNT(25) KCACHE0(cb2, 0 to 15)
    0 x: DOT4            ____, R2.x, 1.0f
      y: DOT4            ____, R2.y, 1.0f
      z: DOT4            ____, R2.z, 1.0f
      w: DOT4            ____, (0x80000000, -0), 0.0f
      t: MULLO_INT       R3.y, R1.x, (0x00000003, 4.2039e-45) SCL_210

    1 x: ADD_INT         R0.x, PS0, (0x00000002, 2.8026e-45)
      y: ADD             R15.y, -PV0.x, 1.0f CLAMP
      z: MUL             R15.z, KCACHE0[7].w, 1.0f
      w: ADD_INT         R2.w, PS0, 1
      t: MULLO_INT       R0.z, R1.y, (0x00000003, 4.2039e-45) SCL_210

    2 x: MUL             R15.x, KCACHE0[8].w, 1.0f
      y: ADD_INT         R0.y, PS1, 1
      z: MUL             R16.z, KCACHE0[9].w, 1.0f
      w: ADD_INT         R0.w, PS1, (0x00000002, 2.8026e-45)
      t: MULLO_INT       R3.x, R1.z, (0x00000003, 4.2039e-45) SCL_210

    3 x: MOV             R1.x, 0.0f
      y: ADD_INT         R1.y, PS2, (0x00000002, 2.8026e-45)
      z: ADD_INT         R1.z, PS2, 1
      w: MOV             R1.w, (0x3F800000, 1)
      t: MULLO_INT       R3.z, R1.w, (0x00000003, 4.2039e-45) SCL_210

02 TEX ADDR(192) CNT(8)

[...]

PR #242, opened 2016-08-04
Implement transform feedback (GX2*StreamOut* functions).

achurch:
See issue #234.


src/libdecaf/src/modules/gx2/gx2_shaders.cpp
```diff
@@ -579,6 +601,54 @@ GX2SetStreamOutEnable(BOOL enable)
...
+      break;
+   case GX2StreamOutContextMode::FromStart:
+      control = control
+         .OFFSET_SOURCE().set(pm4::STRMOUT_OFFSET_FROM_PACKET);
+      srcLo = 0u;
+      break;
+   case GX2StreamOutContextMode::FromOffset:
+      control = control
+         .OFFSET_SOURCE().set(pm4::STRMOUT_OFFSET_FROM_PACKET);
+      srcLo = 0u;  // TODO: where does the offset come from?
```
exjam: BTW it definitely does reuse stream here as the offset, just do srcLo = mem::untranslate(stream)

src/libdecaf/src/gpu/pm4.h
```diff
@@ -718,6 +718,54 @@ struct EventWriteEOP
...
+   STRMOUT_OFFSET_FROM_PACKET           = 0,
+   STRMOUT_OFFSET_FROM_VGT_FILLED_SIZE  = 1,
+   STRMOUT_OFFSET_FROM_MEM              = 2,
+   STRMOUT_OFFSET_NONE                  = 3,
+};
+
+BITFIELD(SBU_CONTROL, uint32_t)
+   BITFIELD_ENTRY(0, 1, uint32_t, STORE_BUFFER_FILLED_SIZE);
+   BITFIELD_ENTRY(1, 2, STRMOUT_OFFSET_SOURCE, OFFSET_SOURCE);
+   BITFIELD_ENTRY(8, 2, bool, SELECT_BUFFER);
```
exjam: This is a 2 bit field, change type from bool to uint8_t

exjam:
Also could you rebase on top of this change I just pushed to master d92925d6622724dfc37247f34b40eea234049a28 ?

Fairly simple change, it just changes bitfields so:
bitfield.field().get() becomes bitfield.field()
bitfield.field().set(x) becomes bitfield.field(x)

src/libdecaf/src/gpu/pm4.h
achurch: buuuh, I is stupid.
Also made STORE a bool while I was at it.

src/libdecaf/src/modules/gx2/gx2_shaders.cpp
achurch: Hmm, if that's the case then the offset mode needs to be something else, because FROM_PACKET would try to use the memory address as an offset into the buffer (which is of course going to overflow). Can you doublecheck, or maybe put up the disassembly somewhere so I can take a look at it?

It could of course be a bug in GX2 itself...

achurch:
Rebase done. I haven't actually tried building it yet, will look at the build output and fix anything that needs it.

exjam:
Can't we just map a stream out context to an OpenGL transform feedback object so we can avoid the whole glBeginQuery, glBeginTransformFeedback, glEndTransformFeedback, glEndQuery, glGetQueryObjectuiv around every single draw call? Because this series of functions is forcing the cpu & gpu to sync in order to get the query result right?

achurch:
Rebased again, and added a cache on the buffer enable bits.

Re transform feedback objects, from what I read they're not a panacea like that -- you still need the Begin/EndTransformFeedback in any case, and XFB objects are only good for caching draw counts if you reuse the count on the GPU side -- you still need a query to retrieve it on the CPU side.

That said, there are probably some cases we could detect and optimize where the client code passes the feedback buffer to a shader, and in that case XFB objects would be a win.

PR #243, opened 2016-08-04
Implement GX2Query{Begin,End} and the OpenGL side of occlusion queries.

achurch:
@exjam This is built on the query stuff you reversed, thanks for that!

(And now, I sleep.)


exjam:
8c09b7c481b13e848afbc6a1d6fe761ee9698363

PR #250, opened 2016-08-05
opengl: Cache current graphics state to avoid redundant GL calls.

achurch:
This roughly doubles frame rate and cuts the number of calls per frame by a factor of 15 on the Xenoblade title and options screens.


exjam:
Never realised drivers were so bad at this!

achurch:
It's not so much that the driver is bad as that a function call into a dynamic library is always going to have lots more overhead than an inline integer comparison or three. Though I imagine modern GL drivers probably have extra layers of overhead from hardware-specific function tables and whatnot.

PR #254, opened 2016-08-06
decaf-sdl: Rewrite emulated screen position.

degasus:
The new code uses integer math to ensure that no pixel will ever be dropped,
and at the same time, no useless bar is shown.
It also fixes an assert error because of floating point math, and the worse
one, because of height >> width.
There is a new parameter for an additional border around both screens, but
it's main purpose is to test this code with eg a single pixel line.

No idea about the exact coding style, but I've also dropped "auto". It was
a mixed usage of float and doubles, hard to check which one is of which format.

What do you think about the design ideas:
- aspect is more important than 5px border between TV and DRC
- symmetry is more important than aspect


brett19:
lgtm 👍 

P.S. Someone really needs to get around to building a Qt UI.

PR #256, opened 2016-08-09
opengl: Handle color buffer write masking with CB_SPECIAL_DISABLE.

achurch:
Fixes the 4th argument to GX2SetColorControl() having no effect (Xenoblade menus use this to suppress color buffer writes while drawing stencil rectangles).


exjam:
Can you change
`gLog->debug("Draw call with unhandled CB_SPECIAL_OP {}", special_op);`
to
`decaf_abort(fmt::format("Draw call with unhandled CB_SPECIAL_OP {}", special_op));`

We are trying to be more strict about catching unimplemented graphics things now since they are very easy to miss in logs!

achurch:
Sure thing, done!

PR #260, opened 2016-08-09
opengl: Fix handling of integer vertex attributes.

achurch:
I realize the bitcasts come from the fact that both int and float instructions use the same GPRs, but I still feel that there ought to be a cleaner way to deal with this...

### DIFF
--- a/src/libdecaf/src/gpu/opengl/opengl_shader.cpp
+++ b/src/libdecaf/src/gpu/opengl/opengl_shader.cpp
@@ -255,7 +255,12 @@ bool GLDriver::checkActiveShader()
                uint32_t divisor = 0;
 
                gl::glEnableVertexArrayAttrib(fetchShader.object, attrib.location);
-               gl::glVertexArrayAttribFormat(fetchShader.object, attrib.location, components, type, normalise, attrib.offset);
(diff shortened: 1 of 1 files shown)


achurch:
Closing for the moment because this maaaay have been correct as written. I've got a feedback shader that didn't give correct outputs originally, but it still doesn't give correct outputs with this change, and the uint shader input has the bit pattern of a float when I write it to an XFB buffer so... I dunno. Will investigate more.

exjam:
To avoid the float / int bit cast we'd have to do liveness analysis on the GPRs to figure out when the contents are integer and when they are float. It is pretty easy to do because integer instructions are explicitly used so you can know what type is in a register just by inspecting the instructions which use it.

It's just we're trying to keep the shader decompiler as simple as possible right now so we avoid doing that for now.

achurch:
Yeah, fair enough. I guess making it work nicely can wait until it works in the first place. (:

exjam:
Best way would be to use Vulkan and can translate to the SSA Spir-V which makes everything very easy for us :+1:

achurch:
Figured it out -- the bitcast is necessary, but we also have to tell OpenGL not to convert to float. Will update the change in a moment.

exjam:
but... why.... why does glVertexArrayAttribIFormat exist?????

PR #261, opened 2016-08-10
opengl: Ensure writes are not masked when clearing buffers.

achurch:
According to apitrace, this fixes Xenoblade model rendering with respect to depth buffers, though it's still not enough to get models actually drawn to the screen.

I didn't attempt to restore the color mask (as I did with the depth mask) because it gets fairly complex due to the interaction between per-buffer masks and the global CB_COLOR_CONTROL mask.


exjam:
Yup that's fine, I have no problem with setting mask to 0xF and letting the next draw reset it, seems like a good idea - maybe we should follow that pattern everywhere.

PR #263, opened 2016-08-10
opengl: Fix failure to reuse color buffers as textures.

achurch:
Fixes Xenoblade 3D rendering not appearing at all. The title screen now looks more or less correct; the character creation screen goes crazy, but at least it shows something.

And yes, Xenoblade renders its world in 768x620. Why 620, I have no clue.


brett19:
I expect that align_up(..., 16) is slightly incorrect.  It should probably be based on the pitch alignment used by the latte GPU (which is based on surface format?).

achurch:
Yeah, I'm not particularly happy with the magic constant 16 in there (esp. since the tile size is 8x8), but I don't understand the GPU well enough yet to suggest anything better.

achurch:
Will rewrite this to resize the texture when needed, as we discussed.

PR #264, opened 2016-08-11
opengl: Use common logic for uniform, attribute, and feedback buffers.

achurch:
This gives me a solid ~30% speed increase in Xenoblade.

### DIFF
--- a/src/libdecaf/src/gpu/opengl/opengl_draw.cpp
+++ b/src/libdecaf/src/gpu/opengl/opengl_draw.cpp
@@ -405,8 +405,6 @@ GLDriver::streamOutBufferUpdate(const pm4::StreamOutBufferUpdate &data)
    auto bufferIndex = data.control.SELECT_BUFFER();
 
    if (data.control.STORE_BUFFER_FILLED_SIZE()) {
-      copyFeedbackBuffer(bufferIndex);
-
       auto addr = data.dstLo;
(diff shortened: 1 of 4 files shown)


achurch:
Rebased on top of the locking changes.

exjam:
Sorry can you rebase this again please :fire: ! Thanks :+1:

brett19:
Could you also split this into two PRs?  We want to accept transform feedback, but need to do a little work on the other part.

exjam:
Actually on further though invalidating tiling apertures is going to require some more work because of how we implement it.

On hardware a tiling aperture is a virtual memory map over the same physical memory of the tiled texture. That is impossible for us to implement so we hack around it by copying and untiling the texture to tiling aperture memory on GX2AllocateTilingApertureEx, and copying and retiling it back to the original memory on GX2FreeTilingAperture. This will not work if the GPU picks up the invalidate before the original texture memory has been updated by GX2FreeTilingAperture.  Or for example if a game never actually frees an aperture and just keeps reusing it.

Really the best fix would to be to forward through the untiled aperture memory all the way through to the GPU so we don't have to do the retile -> untile -> upload stage.

I think for now maybe just remove the tiling aperture invalidation code from this PR and we can revisit it at a later point.

achurch:
Rebased and removed aperture stuff, PTAL.

PR #266, opened 2016-08-11
opengl: Fix failure to reuse color buffers as textures.

achurch:
Rewrite of #263 to resize OpenGL surfaces as necessary. Including depth in the scale vector is probably overkill since this should never be an issue for anything except 2D textures, but I figure better to keep it simple for now and worry about optimization later.


brett19:
achurch: If you have a chance, pop on IRC.

PR #268, opened 2016-08-12
Added a --display-mode option

Adubbz:
Allows switching between windowed split (DRC and TV), windowed-full (just the TV) and fullscreen (also just the TV) modes.

### DIFF
--- a/src/decaf-sdl/decafsdl.cpp
+++ b/src/decaf-sdl/decafsdl.cpp
@@ -60,6 +60,10 @@ DecafSDL::createWindow()
       return false;
    }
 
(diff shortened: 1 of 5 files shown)


exjam:
I merged this as 796fff3af8c1bafdfeffd3b5299095f47ec081da but with a few changes:

1. Moved the config options to the decaf-sdl config files, as they are specific to decaf
2. Split it into 2 options, --display-mode {Windowed/Fullscreen} and --display-layout {Toggle/Split}
3. Made the toggle display layout so you can press Tab to toggle between the TV and DRC screen rather than showing only the TV screen

PR #269, opened 2016-08-12
opengl: Simplify locking.

achurch:


brett19:
This removes one of the locks entirely, which leaves a bunch of shared state without any kind of mutex protecting it.

achurch:
What part of the state is actually shared, except the new task stuff?

brett19:
All of the state inside of the OpenglDriver

achurch:
That's not an answer. :P All the interesting stuff happens via PM4 commands which is all on the same thread, so there's nothing it needs to be locked against.

brett19:
The list of buffers for instance.  The specific case thats going to be relevant is knowing the information about a surface to decide whether we need to do a texture download from copySurface.

PR #271, opened 2016-08-12
opengl: Align texture addresses to a multiple of 0x800.

achurch:


achurch:
Turns out this breaks Shovel Knight, so we apparently do need to take the surface alignment into account.

PR #272, opened 2016-08-12
opengl: Take the absolute value of LOG_* arguments.

achurch:
Xenoblade generates a texture with a pixel shader that uses LOG_CLAMPED instructions; without the abs(), the alpha channel ends up as all NaNs.

The R600 docs suggest that LOG_CLAMPED and LOG_IEEE behave the same way, which makes no sense to me but I've applied abs() to both instructions on that basis.

### DIFF
--- a/src/libdecaf/src/gpu/opengl/glsl2_alu_op2.cpp
+++ b/src/libdecaf/src/gpu/opengl/glsl2_alu_op2.cpp
@@ -346,7 +346,16 @@ LSHR(State &state, const ControlFlowInst &cf, const AluInst &alu)
 static void
 LOG(State &state, const ControlFlowInst &cf, const AluInst &alu)
 {
-   unaryFunction(state, cf, alu, "log2");
+   insertLineStart(state);
+   insertDestBegin(state, cf, alu, state.unit);
+
+   state.out << "log2(abs(";
+   insertSource0(state, state.out, cf, alu);
+   state.out << "))";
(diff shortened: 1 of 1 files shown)


exjam:
This seems a bit odd because the op2.SRC0_ABS exists which applies abs to src0, do you think the NaN could be generated because we do not handle the special cases as per AMD doc:
```
LOG_CLAMPED - Scalar base-2 log.
If (src0 == 1.0f) {
   dst = 0.0f;
} Else {
   dst = LOG_IEEE(src0)
   // clamp dst
   if (dst == -INFINITY) {
      dst = -MAX_FLOAT;
   }
}

LOG_IEEE - Scalar Base-2 IEEE log.
If (src0 == 1.0f) {
   dst = 0.0f;
} Else {
   dst = ApproximateLog2(src0);
}
```

achurch:
The special cases don't apply here -- the NaNs are generated because the input is negative.

achurch:
FWIW, here's what the generated texture looks like with log(abs(...)): http://achurch.org/logtex.png
It seems sensible, which is what makes me think the sign bit on log inputs is ignored.

Regarding SRC0_ABS, that could really go both ways -- if the sign bit is known to be ignored, the shader author might just not bother explicitly taking the absolute value.

exjam:
e8203f1768d31637b8048d3fe79d3559729e25c8

PR #273, opened 2016-08-12
Revert 92e56de and 273871e.

achurch:
gx2.rpl uses PM4 commands for invalidation, so we should do so as well.


brett19:
gx2.rpl does not JUST use a pm4 command.  In the case of CPU invalidations, the operation is synchronous.

exjam:
GX2Invalidate does JUST use a pm4 command! :tada:

PR #274, opened 2016-08-13
Logging fixes.

achurch:
Mainly to avoid the case where the config file is invalid, but the program doesn't print any error (because the default is not to log to stdout) and I get confused...


brett19:
lgtm 👍 ,  I didn't realize that the cli log was something different than the main log.  Why the differentiation?

achurch:
I assume the intent is that gCliLog is used for startup messages, while gLog is used for logging related to the actual emulation. But I haven't looked at the full history of the code, so that's only a guess.

exjam:
Yes what achurch said, due to the fact that the gLog is setup by decaf and prints thread id and shit

PR #276, opened 2016-08-13
gpu: Shift depth buffer addresses by 8 in PM4 packet.

achurch:
As can be seen from GX2SetDepthBuffer().


exjam:
Actually I just checked with gx2.rpl and it's a bug in GX2SetDepthBuffer, we should be shifting by 8 in there, you can modify this PR or i can go push the fix don't mind :)!

```
   uint32_t values2[] = {
      depthBuffer->surface.image.getAddress() >> 8,
      db_depth_info.value,
      depthBuffer->hiZPtr.getAddress() >> 8,
  };
```

achurch:
Hah, fair enough (: Done here.

PR #277, opened 2016-08-13
opengl: Fix overly aggressive texture upload avoidance.

achurch:
I got slightly overenthusiastic with the caching; we can't skip uploading a dirty texture just because the registers haven't changed. (But we can still skip the GL state updates.)

### DIFF
--- a/src/libdecaf/src/gpu/opengl/opengl_driver.h
+++ b/src/libdecaf/src/gpu/opengl/opengl_driver.h
@@ -175,14 +175,8 @@ struct DepthBufferCache
 
 struct TextureCache
 {
-   uint32_t baseAddress = 0;
-   uint32_t word0 = 0;
-   uint32_t word1 = 0;
-   uint32_t word2 = 0;
-   uint32_t word3 = 0;
+   uint32_t surfaceObject = 0;
    uint32_t word4 = 0;
-   uint32_t word5 = 0;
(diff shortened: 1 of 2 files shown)


exjam:
FYI we still have something going wrong, not sure what yet. This was rendering fine last night for me! Maybe its the attribute buffer uploads.

http://gfycat.com/LiveSolidFantail

exjam:
The main menu of Xenoblade has regressed a bit as it doesn't render much of the background scene much anymore, Shovel Knight doesn't render anything in game, and in NES Remix the score counter stays at 0 :-1: :sob:

achurch:
Xenoblade main menu works as well for me as it ever has: (modulo slowness which is due to the CPU load of hashing -- I'm going to add an option to detect static buffers, but that's a separate thing)
xbmenu

I'm busy with other stuff at the moment, but FYI I've put all my local changes on https://github.com/achurch/decaf-emu/tree/local in case there's something there that matters.

brett19:
Why is there a large CPU load of hashing?  Or rather, why is the game invalidating memory its not changed (since we only hash when there is an invalidation).

achurch:
The large CPU load is because the game calls GX2Invalidate all over the place. The game does that because... I dunno, but if I had to guess I'd say it's low overhead on real hardware so they just stuck lots of them in the core draw routines to be safe.

PR #278, opened 2016-08-13
opengl: Fix failure to upload new data when expanding a data buffer.

achurch:
Oops.

Also, I noticed Xenoblade reuses the same portion of memory for different sized attribute buffers (as I imagine a lot of games might do), so we could end up with many GL data buffers overlapping the same address. Not sure offhand what the best solution to this is.


brett19:
I think that this can probably solved just through general GL object garbage collection.  I don't suspect there is any way we can accurately determine the true bounds of the attribute buffer when the games only give us a slice at a time.  Though perhaps at the very least we might be able to mark chains of buffers as being linked together, so at some point they can be reallocated as a single big opengl buffer.  Though this could cause size-creep, and the object would never be GC'd because a small fraction of it is still in use.

brett19:
Keep in mind this is just a memory problem, correctness is still maintained even though many GL buffers reference the same PPC memory.

PR #290, opened 2016-08-16
gx2: Use correct frame rate for vsync handler.

achurch:


brett19:
Let's stick with the /60 rate for now (due to rounding, it might actually be close to 59.xx anyways).  Note that in the future we should probably sync up with the host's vsync rate when possible, otherwise we are going to have some funny results (though if the host refresh rate is not within 1fps of 59.xx fps, we should probably fall back to our own clock).

PR #295, opened 2016-08-16
Honor the swap interval set by the game.

achurch:
The delay in decafSwapBuffers() is intentionally unconditional on whether the host GPU has v-sync enabled. If the host frame rate is less than 59.94 fps, it'll just sync to that rate; if it's greater than 59.94, there will be an occasional 1-frame delay, but you'd still get an effective 1-frame delay at the same interval without host v-sync (you'd just see tearing instead).

I added the [59,61] range check on host frame rate per Brett's suggestion, though I tend to think an explicit config option for enabling host v-sync would be more appropriate.

Note that I don't really grok std::chrono, so there may be a better way to write the decafSwapBuffers() delay logic.

(This PR includes #293 since GitHub doesn't seem to allow a PR comparing between two branches of a different fork.)

### DIFF
--- a/src/decaf-sdl/decafsdl.cpp
+++ b/src/decaf-sdl/decafsdl.cpp
@@ -144,10 +144,19 @@ DecafSDL::run(const std::string &gamePath)
             mGraphicsDriver->run();
          } };
    } else {
-      // Set the swap interval to 0 so that we don't slow
-      //  down the GPU system when presenting...  The game should
-      //  throttle our swapping automatically anyways.
-      SDL_GL_SetSwapInterval(0);
+      // If the host frame rate is reasonably close to 60fps, use host vsync
+      //  when the game requests it; otherwise disable host vsync (GLDriver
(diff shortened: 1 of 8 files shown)


exjam:
Looks like SDL only allows setting of -1, 0, or 1 for SDL_GL_SetSwapInterval. I'm pretty sure GX2SetSwapInterval is supposed to allow 0+, as in like 0 for no vsync, 1 for vsync 60fps, 2 for 30fps, 3 for 20fps, etc...

achurch:
That looks like bad documentation (a hallmark of SDL...) because the actual code just passes the interval straight to the driver, and both eglSwapInterval and glXSwapIntervalEXT do what you'd expect (wait the given number of frames per swap call).

brett19:
I am going to reject this particular PR for now.  There is a lot of known issues with the way we currently handle swapping, and this particular PR potentially could make things worse.  I hope to be able to take a look some point soon.  Maybe this evening.

PR #301, opened 2016-08-17
opengl: Exclude degamma from surface key.

achurch:
This partly fixes the Xenoblade character creation screen, which renders the world to a surface in RGB mode and then reads from the same surface in SRGB mode; the circular floor is now a real floor instead of a black hole. The remaining problem (white glare) seems to be some sort of HDR math glitch, which I'm still looking into.

### DIFF
--- a/src/libdecaf/src/gpu/opengl/opengl_surface.cpp
+++ b/src/libdecaf/src/gpu/opengl/opengl_surface.cpp
@@ -692,7 +692,7 @@ GLDriver::getSurfaceBuffer(ppcaddr_t baseAddress,
    //  are compatible across.  Note that at some point, we may
    //  need to make format not be part of the key as well...
    auto surfaceKey = static_cast<uint64_t>(baseAddress) << 32;
-   surfaceKey ^= format << 22 ^ numFormat << 28 ^ formatComp << 30 ^ degamma << 31;
+   surfaceKey ^= format << 22 ^ numFormat << 28 ^ formatComp << 30;
 
    switch (dim) {
    case latte::SQ_TEX_DIM_1D:


brett19:
Maybe we should drop other parts of the key that are not relevant to the memory layout semantics of the real GPU?

achurch:
I looked at the key and offhand didn't see anything that definitely shouldn't be included. There is an argument for dropping everything and just letting the game do whatever it wants to do with formats, but I haven't gone that far just because I don't have an example yet.

Also, it just occurred to me that the fix here may not be complete -- the GL surface mode is still RGB even if the client requests SRGB, so the sampler won't return the correct values. We may need to treat this in the same manner as a size change, and create another surface in the chain with a glCopyImageSubData call.

exjam:
Yeah I was wondering how its meant to be interpreted when you have an RGB and SRGB texture in the same memory :confused: I guess write as RGB and read as SRGB, which yeah would involve a copy. Although that implies that it's actually going to be 2 GL surfaces and we do indeed need to keep degamma in the shader key!

achurch:
I haven't actually looked closely at @brett19's new surface management code so he might be better placed to make that change -- I do think removing it from the key is correct, since e.g. height is omitted for 2D textures, but there needs to be some extra logic to record whether the current surface is degamma and do a CopyImageSubData to an SRGB slave surface if not. Brett could probably fix that faster than I could. (look at me weaseling out of the hard work, lol)

brett19:
Hey @achurch, I'd be happy to, if you want to do it, it's pretty trivial, just add degamma to the HostSurface properties and add it to the check at the top of getSurfaceBuffer that decides if it can reuse an existing HostSurface or needs to build a new one.

achurch:
I'm kinda focusing on that HDR bug now, so if you could take care of it I'd appreciate it -- thanks!

brett19:
Does this work? https://github.com/decaf-emu/decaf-emu/tree/degammahostsurf

achurch:
Yup, seems to be fine!

PR #307, opened 2016-08-17
opengl: Implement the JUMP shader instruction.

achurch:
Treating it as a no-op and letting the active tests on ALU blocks deal with it only works as long as there are no nested conditionals; we have to honor a JUMP that jumps over a nested JUMP/ELSE.


exjam:
Have you got an example of a shader where this is needed? If a shader uses nested conditionals it should be using PUSH / POP ?

achurch:
I had one which I traced through and it seemed to misbehave, but tracing through it again now, it seems to work fine -- I guess I didn't follow the stack closely enough. Some of the conditional ALU CF instructions don't have pushes, but it looks like they're all innermost blocks, so the lack of a push won't hurt anything.

Fair enough, I retract this PR.

PR #312, opened 2016-08-18
decaf-sdl: Add joystick support.

achurch:
I wanted to save the default controller type in the config file too, but I don't think cereal can deal with enums?

### DIFF
--- a/src/decaf-sdl/config.cpp
+++ b/src/decaf-sdl/config.cpp
@@ -1,3 +1,5 @@
+#include "clilog.h"
+#include "common/decaf_assert.h"
 #include "config.h"
 #include "libdecaf/decaf_config.h"
 #include <climits>
@@ -25,6 +27,51 @@ bool force_sync = false;
 
 } // namespace gpu
 
+namespace input
+{
+
+std::vector<InputDevice> devices;
+
+namespace vpad0
+{
+
+ControllerType type = Keyboard;
+std::string name = "";
+
+} // namespace vpad0
+
+std::string
+controllerTypeToString(ControllerType type)
+{
+   switch (type) {
+   case None:
+      return "none";
+   case Keyboard:
+      return "keyboard";
+   case Joystick:
+      return "joystick";
+   }
+
+   decaf_abort(fmt::format("Invalid controller type {}", type));
+}
+
+ControllerType
+controllerTypeFromString(const std::string &typeStr)
+{
+   if (typeStr.compare("keyboard") == 0) {
+      return Keyboard;
+   } else if (typeStr.compare("joystick") == 0) {
+      return Joystick;
+   } else {
+      if (typeStr.compare("none") != 0) {
+         gCliLog->error("Invalid input type: {}", typeStr);
+      }
+      return None;
+   }
+}
+
+} // namespace input
+
 namespace log
 {
 
@@ -34,40 +81,6 @@ bool to_stdout = false;
 std::string level = "debug";
 
 } // namespace log
-
-namespace input
-{
-
-namespace vpad0
-{
-
-std::string name = "keyboard";
-int button_up = SDL_SCANCODE_UP;
-int button_down = SDL_SCANCODE_DOWN;
-int button_left = SDL_SCANCODE_LEFT;
-int button_right = SDL_SCANCODE_RIGHT;
-int button_a = SDL_SCANCODE_X;
-int button_b = SDL_SCANCODE_Z;
-int button_x = SDL_SCANCODE_S;
-int button_y = SDL_SCANCODE_A;
-int button_trigger_r = SDL_SCANCODE_E;
-int button_trigger_l = SDL_SCANCODE_W;
-int button_trigger_zr = SDL_SCANCODE_R;
-int button_trigger_zl = SDL_SCANCODE_Q;
-int button_stick_l = SDL_SCANCODE_D;
-int button_stick_r = SDL_SCANCODE_C;
-int button_plus = SDL_SCANCODE_1;
-int button_minus = SDL_SCANCODE_2;
-int button_home = SDL_SCANCODE_3;
-int button_sync = SDL_SCANCODE_4;
-int left_stick_x = -1;
-int left_stick_y = -1;
-int right_stick_x = -1;
-int right_stick_y = -1;
-
-} // namespace vpad0
-
-} // namespace input
 
 struct CerealDebugger
 {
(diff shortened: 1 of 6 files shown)


brett19:
Someone needs to PR a better config tool...

exjam:
Well I kind of intended for vpad0::name to indicate what type it was, i.e. name == "keyboard" for keyboard, then name == "x" for joystick and would look up name X in the sdl joystick list.

achurch:
That's overlaying two purposes onto one piece of data, which is bad design IMO. Also, you need different sets of controls for different devices anyway -- separate minus/plus keys for keyboard axis inputs vs. a single axis index for joystick inputs.

[Separate note -- I'm not sure whether Wii U analog sticks have up=-Y and down=+Y or the reverse; the logic there may need to be flipped.]

exjam:
Well technically you need separate minus/plus for analog sticks anyway because you need to say which way a stick points, for example:
left stick up = axis 5 positive
left stick down = axis 5 negative

otherwise you're stuck trying to imply the direction by doing left_stick_v = axis_5_positive!

exjam:
Although all this talking about input configs gets really really confusing when you consider WPAD...

So VPAD is _always_ the DRC controller (the one with the screen), technically the API supports 2 DRC controllers, but I don't think this is valid use case for anything.

With WPAD you can have any of the other Wii controllers, such as a Wiimote, Nunchuk, Classic, Pro Controller, and you can have up to 4 of these controllers in any combination!

So that's why I was trying to divide the input config up to only wpad0, vpad0, vpad1, vpad2, vpad3 and not wanting to make it more complicated than that!

achurch:
Fair point re: axis signs, though I think that doesn't affect the benefit of separate configs -- there would be two pieces of data per axis, yes, but they'd be different ("minus key" and "plus key" vs. "axis index" and "invert flag").

Also, it's perfectly reasonable for the user to want to keep separate joystick and keyboard configs without wanting to maintain entirely separate config files for each.

exjam:
Well I did also have the idea of that instead of mapping controller configs to specific controllers (wpad0, vpad0, vpad1, vpad2, vpad3), that you could define only the input mappings per host controller, and then we could assign the guest controller at run time. Such that you would end up with in your config like "i want keyboard to be a DRC controller with X=KEY_Z etc" or "i want joystick named ps4_controller to be a Wii U Pro Controller with A=Button_9 etc", and then in the emu you just assign players using some yet undetermined magic - like maybe pressing the start button on a controller to set assign it to a guest controller.

achurch:
Yeah, that would make sense. How about something like this:
```
input {
  devices=[{type: "keyboard|joystick|...", name="PS8 TWELVEAXIS", button_a=13, ...}, {...}]
  vpad0 {
    type="keyboard"
    name=""
  }
  wpad0 {
    type="joystick"
    name="PS8 TWELVEAXIS"
  }
}
```
where you first define all the input devices you want, then you assign them to Wii U devices.

I don't actually know if you can do that array of maps thing in cereal, but other than that...

exjam:
Maybe this will finally be the situation to move us to a new config file format... I've heard people suggest YAML, do you know of any config file formats? TBH almost anything would be better than our super strict cereal json parser.

achurch:
I haven't actually seen any programs using YAML before, and a quick look at the syntax makes me think it's not particularly user-friendly to edit -- all the mysterious punctuation feels fragile, in that it could break easily and not be obvious what's broken. Granted, that might not be a huge deal if we had a configuration UI, but we don't have that yet so...

XML isn't great in terms of readability either but at least it's widely known, and if nothing else there's libxml2 to parse it.

Frankly, when I need to use a config file and it's not super complicated I just do a simple INI style thing and write code for it myself, because it's faster each individual time to do that than to try and find a good config library :P

exjam:
Yeah that was my feeling over YAML too, tbh I quite like JSON.

https://xkcd.com/927/ ? :+1:  I mean we could make our own... but it'd end up looking just like a less strict JSON I imagine.

Actually saying that i just googled for a non strict json and this came up https://hjson.org/

achurch:
Haha, I was half tempted to quote that xkcd myself...

Yeah, hjson looks more sensible (which granted is not a high bar to cross at the moment). I guess the issue is how configurable it is, i.e. do we have to specify a static format or will it give us a simple map we can index into.

exjam:
On further inspection it looks like there is no hjson parser library for c / c++ that I can find, so we'd have to write it from scratch using their grammar anyway haha...

exjam:
Right so rapidjson is included within cereal and can actually be used here,
would be `#include <cereal/external/rapidjson/reader.h>` then we use the options: kParseCommentsFlag, kParseTrailingCommasFlag, which makes it a bit less strict.

Then with rapidjson you don't have to specify the format ahead of time - that was just a cereal thing, so you can access key values with just like document["input"]["key_x"] and document.HasMember("input") etc. The only difference between that and a more human friendly format is that key names and string values will still need "" around them, but that's not a deal breaker imo.

achurch:
Oh, nice! Okay, I'll open up a separate PR to change cereal to rapidjson (unless someone beats me to it), then redo the input config like I was suggesting and ping this PR.

Will get to it later once I work through this Xenoblade title screen thing...

exjam:
Yeah I can do it later today if you want, in like 3~4 hours from now when I finish with work :+1:

exjam:
@achurch did you start on moving to new config or shall I do that now?

exjam:
I just realised generating a default config sucks balls without cereal though :-1: 

Will have to make my own wrapper around rapidjson if i want auto read / write ... sigh :sob:

achurch:
Nope, I'm still staring at apitrace. Good luck with the wrapper...

brett19:
@achurch : Any luck with that?  P.S. IRC! :D

exjam:
I decided to just modify cereal to be more lenient instead, 78daf1f1d661424388d8f25e85ed1a48d675c1d7

achurch:
Cools, I'll update this and ping it when the changes are done.

achurch:
@exjam Change updated, PTAL. I ended up needing a fair amount of kludgery: cereal refused to detect out-of-line serializer functions for InputDevice, and when I tried to use the save/load_minimal thing to serialize the ControllerType enum as a string, cereal complained that there was no associated output archive format for JSONOptionalInputArchive -- and when I copied JSONOutputArchive into that header, it proceeded to ignore save/load_minimal and just write the enum as an int. But it all seems to work (knock on wood...)

exjam:
Merged as 4b5208898f1dbfee51386a881266916eb2f5e65b

Though I did have to fix where you used SDL_CONTROLLER_BUTTON_DPAD_UP for a bunch of buttons because of copy paste :laughing:

achurch:
lol, oops... thanks for fixing that (:

PR #314, opened 2016-08-18
GL rendering fixes.

achurch:
With these fixes I'm pretty sure all the binding stuff is now correct. I left a TODO about a case I noticed where there can be lots of unnecessary copies between textures and color buffers at the same address, but I haven't actually fixed it yet.


exjam:
We should probably assert that the other shader register offsets are also 0!

SQ_PGM_CF_OFFSET_ES 0x288d8 
SQ_PGM_CF_OFFSET_FS 0x288dc 
SQ_PGM_CF_OFFSET_GS 0x288d4 
SQ_PGM_CF_OFFSET_PS 0x288cc 
SQ_PGM_CF_OFFSET_VS 0x288d0

achurch:
Good call -- I thought it odd that there was only a VS register in the register list. I'll add the registers and assertions into that commit and FP-update.

PR #316, opened 2016-08-18
opengl: Use DCFlush()/DCStore() as the dirty-resource signal.

achurch:
For well-behaved games, this is sufficient to cover all GPU resource updates from the CPU side (modulo shaders, which I'll PR next). If any games are non-well-behaved and still happen to work on real hardware, that's probably the sort of thing that has to be handled on a case-by-case basis, like race conditions that happen to never be hit on a real 750CL or such...

### DIFF
--- a/src/libdecaf/decaf_graphics.h
+++ b/src/libdecaf/decaf_graphics.h
@@ -15,6 +15,8 @@ public:
    virtual void run() = 0;
    virtual void stop() = 0;
    virtual float getAverageFPS() = 0;
+
+   virtual void handleDCFlush(uint32_t addr, uint32_t size) = 0;  // May be called from any thread!
 };
 
 class OpenGLDriver : public GraphicsDriver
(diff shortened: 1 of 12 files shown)


brett19:
@exjam and I have discussed this a bit, I'll let him comment.

exjam:
Is it safe to have std::map<x, Type> map; then do {lock ptr = &map[x]},  then use pointer after unlock? Doesn't that heavily rely on how the map's underlying storage function? Like you couldn't do that with vector for example.

I do agree though that watching for a CPU flush is the correct way to dirty buffers.

achurch:
I haven't read the actual STL spec, but online docs I've seen say pointers are guaranteed to be stable, e.g. http://en.cppreference.com/w/cpp/container/unordered_map

brett19:
We only need to be careful when deleting an item that we hold a reference to still.  Deleting Item A does not invalidate Item B.  This is the same behaviour as std::list.

achurch:
Xenoblade was using LCStoreDMABlocks() to copy feedback buffers around. With this change, the Xenoblade title screen isn't perfect (note the sharp lines in the cloud textures) but it's close:
xbtitle

JMC47:
In my expert opinion, this makes a game look better therefore it must be correct.  And people wonder why I'm a terrible tester.

exjam:
Merged.
51e7707e74fdf702b690120203517dc1d21bdad3
0c4797f2d985f8d55fff34c87553ab9398310b50

PR #319, opened 2016-08-18
opengl: Invert handling of DX_CLIP_SPACE_DEF in PA_CL_CLIP_CNTL.

achurch:
The OpenGL API docs for glClipControl explicitly give the example of DirectX as a use case for GL_ZERO_TO_ONE, and Mesa includes the expression S_028810_DX_CLIP_SPACE_DEF(state->clip_halfz) when computing the PA_CL_CLIP_CNTL register value (note the "halfz"), so I suspect the current logic is backwards.

### DIFF
--- a/src/libdecaf/src/gpu/opengl/opengl_registers.cpp
+++ b/src/libdecaf/src/gpu/opengl/opengl_registers.cpp
@@ -262,9 +262,9 @@ GLDriver::applyRegister(latte::Register reg)
       }
 
       if (pa_cl_clip_cntl.DX_CLIP_SPACE_DEF()) {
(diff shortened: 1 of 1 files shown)


exjam:
Not sure which change did it @achurch but you is my hero https://images-2.discordapp.net/.eJwNyMENwyAMAMBdGAALHLsl2yCCSNQEI-yoj6q7t_e8j7vn6Va3mw1dAbZDi8zNq8nMrfom0s6ax6G-yAXZLJf9qt0UYsCICR-BIxJj5OVflAgXJuJnCokI4e6vLu_uR2_u-wMG6SLj.tJnmB4YoO4EUqjVZlSJ2MT_oif8.png

PR #322, opened 2016-08-19
opengl: Avoid repetitive glEnable(GL_PRIMITIVE_RESTART) calls.

achurch:
If it is permanently enabled, there's no reason to re-enable the GL state on every register write.

Could also be that the register just has a default value of 1 instead of 0, but I don't have a good way to test graphics code on hardware at the moment.

### DIFF
--- a/src/libdecaf/src/gpu/opengl/opengl_driver.cpp
+++ b/src/libdecaf/src/gpu/opengl/opengl_driver.cpp
@@ -34,6 +34,9 @@ GLDriver::initGL()
    mActiveShader = nullptr;
    mDrawBuffers.fill(gl::GL_NONE);
 
+   // Primitive restart appears to always be enabled on the GX2
+   gl::glEnable(gl::GL_PRIMITIVE_RESTART);
+
    // Create our blit framebuffer
    gl::glCreateFramebuffers(2, mBlitFrameBuffers);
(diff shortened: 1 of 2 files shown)


exjam:
These were the intial values that I dumped out of gx2.rpl:

https://github.com/decaf-emu/decaf-emu/blob/master/src/libdecaf/src/modules/gx2/gx2_contextstate.cpp#L16-L112

exjam:
Actually, prim reset is there... I must have a bug in the code which loads this default state then...
VGT_MULTI_PRIM_IB_RESET_EN       = 0x28A94,

So (0x28A94 - 0x28000) / 4 = 0x2a5

And in the list you can see
{ 0x2A5, 1 },

exjam:
Oh wait, that doesn't set the register to value 1 though, this is just the list of what registers to actually load from the context state, I must have missed where it initialises the register value.

exjam:
OK OK OK sorry for spam everyone, I found where it sets it, we definitely should be using the _EN register to enable it in the OpenGL side. But it's going to take me a while to reverse all the initialised registers here so gimme some time before I get that change up :+1:

achurch:
Cools -- yeah, seemed odd to me that the Wii U would get a special version of the GPU with that register hardwired to 1 (: Closing this since it's no longer relevant.

PR #327, opened 2016-08-20
Misc Tweaks

Adubbz:
Fixed fullscreen borders, escape now exits decaf, added --stretch (disabled by default)


exjam:
Merged with a few tweaks.

Renamed --stretch to --display-stretch.
Split into 3 commits, because there was 3 separate changes here!

PR #332, opened 2016-08-20
opengl: Disable early fragment tests for shaders which use discard.

### LINKED ISSUE #331: Several games have a broken alpha on textures
There are plenty of games with a broken alpha caused by https://github.com/decaf-emu/decaf-emu/pull/314

ice

There is a **[video](https://www.youtube.com/watch?v=vnfM4nBtqN0)** of some 

@achurch ping?

LOG - http://pasted.co/1ebb8a2a

**Edit:**
I found that it comes from https://github.com/decaf-emu/decaf-emu/pull/314/commits/9cefb15af1f8a73825b63deae57942ca2a3697bf

and after revert it's ok again 

ice 2

achurch:
I suspect the way the hardware works wrt early-Z and discard is to disable early-Z whenever the KILL_ENABLE bit is set, but since we know for certain whether the fragment shader calls discard, I've used that knowledge instead. (I don't actually know what the GPU would do if a pixel shader has KILL instructions but the program doesn't set the KILL_ENABLE bit, but I imagine dev-side shader compilers take care of avoiding that.)

Fixes #331.

### DIFF
--- a/src/libdecaf/src/gpu/opengl/glsl2_alu_op2.cpp
+++ b/src/libdecaf/src/gpu/opengl/glsl2_alu_op2.cpp
@@ -271,6 +271,8 @@ binaryCompareKill(State &state,
    insertLineStart(state);
    state.out << '}';
    insertLineEnd(state);
+
+   state.shader->usesDiscard = true;
 }
 
 static void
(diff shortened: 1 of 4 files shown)


brett19:
I think it would be better to follow the hardware, and use the KILL_ENABLE bit to decide (I wonder if I can find a way to confirm thats what the hardware does).

achurch:
Fair enough, done. I've left in discard detection and used it to check that shaders really do always set KILL_ENABLE when using KILL instructions.

brett19:
Perfect, exactly the kind of logic I like to see.

PR #334, opened 2016-08-21
gpu: Remove flags from EVENT_WRITE high address word.

achurch:
Fixes dynamic objects on Xenoblade title screen not being rendered.

FWIW, Mesa's R600 driver suggests that addrHi has 16 valid bits of address, based on code like:
```
radeon_emit(cs, PKT3(PKT3_EVENT_WRITE, 2, 0));
radeon_emit(cs, EVENT_TYPE(EVENT_TYPE_ZPASS_DONE) | EVENT_INDEX(1));
radeon_emit(cs, va);
radeon_emit(cs, (va >> 32) & 0xFFFF);
```
But I don't know where the 8-bit field width came from, and it'll always be zero on the Wii U anyway, so I haven't touched that.

### DIFF
--- a/src/libdecaf/src/gpu/opengl/opengl_driver.cpp
+++ b/src/libdecaf/src/gpu/opengl/opengl_driver.cpp
@@ -441,8 +441,7 @@ GLDriver::eventWrite(const pm4::EventWrite &data)
       decaf_abort(fmt::format("Unexpected event type {}", type));
    }
 
-   switch (data.addrLo.ENDIAN_SWAP())
-   {
+   switch (data.addrLo.ENDIAN_SWAP()) {
    case latte::CB_ENDIAN_NONE:
       break;
    case latte::CB_ENDIAN_8IN64:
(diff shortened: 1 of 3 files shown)


exjam:
Yup looks like I was wrong to copy&paste from event write EOP to event write!

src/libdecaf/src/gpu/opengl/opengl_driver.cpp
```diff
@@ -441,8 +441,7 @@ GLDriver::eventWrite(const pm4::EventWrite &data)
       decaf_abort(fmt::format("Unexpected event type {}", type));
    }
 
-   switch (data.addrLo.ENDIAN_SWAP())
-   {
+   switch (mPendingEOP.addrLo.ENDIAN_SWAP()) {
```
exjam: Why are we using mPendingEOP's ENDIAN_SWAP here? This doesn't feel correct

achurch: Because I accidentally deleted too much code then undeleted it from the wrong place. Derp.

Issue #311, opened 2016-08-18
Change --log-file to allow specifying the log filename.


ttay24:
So, I've looked at the log options and I'm not sure if it's brokey broke or if i'm doing it wrong.

I've tried the following:
> --log-file "logs/log.txt"
> --log-file "logs/"

But neither seem to redirect any of the logs (the main log or the log for a game). Any guidance here?

exjam:
Unfortunately the --log-file argument is a bit misleading here, we should probably change it.

Currently --log-file just enables logging to file, rather than specifying a path for the file.

https://github.com/decaf-emu/decaf-emu/blob/master/src/decaf-sdl/main.cpp#L165-L167

ttay24:
Ah, I see.

So would we want this to be split into two args? One for logging on/off, and another to specify a directory? Would the directory be used for the main log file and any log files for games?

exjam:
2d1252db53fd23f0adbb3526463250194c25c69e

PR #339, opened 2016-08-22
config: added option --log-dir to specify output directory for log (#311)

ttay24:
Based on #311


exjam:
Merged as 2d1252db53fd23f0adbb3526463250194c25c69e

PR #344, opened 2016-08-22
opengl: Only upload uniform registers which have been recently written.

achurch:
This makes apitrace a _lot_ less painful...


exjam:
Is this beneficial for the general case or only for apitrace? If it's just for apitrace we should put it behind a bool like the other debug options.

achurch:
Yes, it should be beneficial in general, though not to the same degree -- I suspect apitrace is doing a lot of un-/serialization of float arrays that slows down large glUniform calls. I certainly can't imagine any (non-contrived) case in which it would reduce performance.

PR #346, opened 2016-08-22
opengl: Detect changes to shader memory and recreate as appropriate.

achurch:
Adds garbage collection of orphaned shader programs as a side effect (because it wasn't really feasible to do this correctly otherwise).


exjam:
Is a good change, but is going to majorly clash with geometry shaders. So i'll let Brett deal with this... :)

brett19:
Yea... There are about to be lots of major conflicts with Geometry Shaders.  Even with my own work I need to do on shaders.  I'll look at this later when I'm at my computer, and probably just merge it ahead of Geom Shaders.

PR #348, opened 2016-08-22
opengl: Cache additional GL state data.

achurch:
FYI, this gives me a performance boost of about 15% on the Xenoblade title screen.

(One of these days I just know I'm going to type Xenogears instead of Xenoblade...)


achurch:
Added suppression of glCheckFramebufferStatus() when unnecessary (per the GL spec). The additional improvement isn't a lot, but I do see FPS peak a bit higher than before so probably still worth it.

brett19:
Conflicts now.  Needs a rebase.

achurch:
Rebased.

achurch:
Rebased again.

PR #356, opened 2016-08-23
swkbd: Fix format of string returned by GetInputFormString().

achurch:
Fixes Xenoblade name entry getting garbage characters.

### DIFF
--- a/src/libdecaf/src/modules/swkbd/swkbd_core.cpp
+++ b/src/libdecaf/src/modules/swkbd/swkbd_core.cpp
@@ -122,13 +122,15 @@ GetDrawStringInfo(DrawStringInfo *info)
 }
 
 
-const wchar_t *
+const be_val<uint16_t> *
 GetInputFormString()
 {
    std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>> converter;
    auto wide = converter.from_bytes(sInputBuffer);
-   auto result = reinterpret_cast<wchar_t *>(sWorkMemory);
-   std::memcpy(result, wide.data(), wide.size() * 2);
+   auto result = reinterpret_cast<be_val<uint16_t> *>(sWorkMemory);
+   for (auto i = 0; i < wide.size(); ++i) {
+      result[i] = static_cast<uint16_t>(wide[i]);
+   }
    result[wide.size()] = 0;
    return result;
 }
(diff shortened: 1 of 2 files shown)


exjam:
We should be using char16_t!

http://en.cppreference.com/w/cpp/language/types

"char16_t - type for UTF-16 character representation, required to be large enough to represent any UTF-16 code unit (16 bits). It has the same size, signedness, and alignment as std::uint_least16_t, but is a distinct type."

achurch:
I tried char16_t in the template but got compiler errors. If you can find a way around that, by all means use it.

We certainly _shouldn't_ use char16_t in the return type because, like uint_least16_t, it's allowed to be wider than 16 bits.

exjam:
That's odd, http://en.cppreference.com/w/cpp/locale/codecvt_utf8_utf16 says "Elem - either char16_t, char32_t, or wchar_t", but yes I agree with you for the return type at least!

achurch:
FWIW, here's the error:
```
/usr/bin/../include/c++/v1/locale:3888:43: error: cannot initialize a parameter
      of type 'intern_type *' (aka 'char16_t *') with an lvalue of type
      'wchar_t *'
                                          __to, __to_end, __to_nxt);
                                          ^~~~
/usr/bin/../include/c++/v1/locale:3799:17: note: in instantiation of member
      function 'std::__1::wstring_convert<std::__1::codecvt_utf8_utf16<char16_t,
      1114111, 0>, wchar_t, std::__1::allocator<wchar_t>,
      std::__1::allocator<char> >::from_bytes' requested here
        {return from_bytes(__str.data(), __str.data() + __str.size());}
                ^
src/libdecaf/src/modules/swkbd/swkbd_core.cpp:129:26: note: 
      in instantiation of member function
      'std::__1::wstring_convert<std::__1::codecvt_utf8_utf16<char16_t, 1114111,
      0>, wchar_t, std::__1::allocator<wchar_t>, std::__1::allocator<char>
      >::from_bytes' requested here
   auto wide = converter.from_bytes(sInputBuffer);
                         ^
/usr/bin/../include/c++/v1/__locale:1022:28: note: passing argument to parameter
      '__to' here
              intern_type* __to, intern_type* __to_end, intern_type*& __...
                           ^
```
It looks like maybe the "w" in "wstring_convert" is the culprit, but there doesn't seem to be a "u16string_convert" to go with u16string, so I dunno. I admit my eyes tend to glaze over whenever I see STL template error spam...

exjam:
Merged as 6bc50065c420565fca044e0e65d248f47e2de6aa

PR #362, opened 2016-08-24
opengl: Use DCInvalidateRange() as a signal to download data buffers.

achurch:
This is the counterpart to the DCFlush change. For this, we do need to synchronize with the GPU thread (the data has to actually be available to the CPU when the DCInvalidate handler returns), so I've added back a simplified version of Brett's task code. I ended up using the same dummy buffer method to pump the PM4 loop if necessary; it would be cleaner to have a shared wakeup signal, but that's more involved and requires changes to the queue handling as well, so I think this is good enough for now.

### DIFF
--- a/src/libdecaf/decaf_graphics.h
+++ b/src/libdecaf/decaf_graphics.h
@@ -16,8 +16,12 @@ public:
    virtual void stop() = 0;
    virtual float getAverageFPS() = 0;
 
-   virtual void notifyCpuFlush(void *ptr, uint32_t size) = 0;  // May be called from any thread!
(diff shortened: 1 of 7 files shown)


exjam:
btw you'd need to rebase on top of this: 05e65fd5cf3143adaf4b4382ef981d0040403135 (I'm trying to get a pm4 capture / replay tool working so you can help us with graphical bugs in other games without owning the games :+1: !)

But also I think @brett19 had some changes to make before merging this, so may want to wait for his feedback.

@brett19 @brett19 @brett19 @brett19 @brett19 @brett19 @brett19 @brett19 @brett19 @brett19

brett19:
This is probably an okay change to include right now.  Note that some games have been having issues with regards to the new invalidation handling we have.

PR #369, opened 2016-08-27
opengl: Don't reject depth+stencil buffers in 24_8 format.

### LINKED ISSUE #330: Stencil buffer binding is incorrect.
Currently, if the incoming buffer is not of the 3 stencil buffer formats, we set stencil_enable to off.  This causes an issue because some games set a format that has room for the stencil, and then expect to use it (there is no difference from the GPUs perspective).  It looks like we need to bind stencil_depth always maybe?  pcmaker dropped the if statement entirely and stuff started working again.

achurch:
This might fix #330.

### DIFF
--- a/src/libdecaf/src/gpu/opengl/opengl_depthbuffer.cpp
+++ b/src/libdecaf/src/gpu/opengl/opengl_depthbuffer.cpp
@@ -36,8 +36,10 @@ bool GLDriver::checkActiveDepthBuffer()
    gl::GLuint surfaceObject = surface ? surface->active->object : 0;
 
    if (format != latte::DEPTH_8_24
-      && format != latte::DEPTH_8_24_FLOAT
-      && format != latte::DEPTH_X24_8_32_FLOAT) {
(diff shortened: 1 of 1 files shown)


exjam:
Those formats don't exist, I think brett was referring to DEPTH_X8_24 and DEPTH_X8_24_FLOAT?

  ```
 if (format != latte::DEPTH_8_24
  && format != latte::DEPTH_8_24_FLOAT
  && format != latte::DEPTH_X8_24
  && format != latte::DEPTH_X8_24_FLOAT
  && format != latte::DEPTH_X24_8_32_FLOAT) {
      // You cannot bind a stencil buffer if the framebuffer does not support it.
      stencil_enable = false;
   }
```

exjam:
I pushed that as e44f92a92b5a4cc766390e3a771a7cddc3131b77

achurch:
Oh wow, apparently my brain confused "FMT" and "DEPTH". Thanks for the fix.

PR #373, opened 2016-08-30
coreinit: Fix uint8_t overflow.

achurch:
Written so 0.0 translates to 128 instead of a potentially unstable value between 127 and 128, though I don't know if that's what hardware does.


exjam:
Nice catch

PR #374, opened 2016-08-31
Fix OpenGL locking errors and other warnings detected by -Wshadow.

achurch:
Been meaning to get to this for a while. The logic errors I spotted were originally mine, so sorry about that. (I'm mildly surprised that it's possible to construct a lock that locks nothing, but whatever...)


exjam:
ecced575ac827fd270ef58761109bfc02bc93811

PR #376, opened 2016-09-02
snd_core: Add a few more AXFX stubs.

achurch:
Xenoblade sets AXFXDelayCallback and AXFXReverbStdCallback as callbacks, so they get called on every audio frame, resulting in massive log spam if they're not stubbed.


brett19:
Lalalala.  Builds so slowww.

brett19:
P.S. You might want to implement the GetMemSize stubs as well.

achurch:
Will do, hang a sec and I'll FP this branch.

PR #387, opened 2016-12-14
opengl: Use a sorted resource map to optimize resource updates.

achurch:
Since resources are not currently expunged when the memory is reused for another purpose, the resource maps grow over time, causing the linear scans in notify{Cpu,Gpu}Flush() and surfaceSync() to consume more and more CPU as time goes on.  This change introduces a ResourceMemoryMap class to keep a sorted list of resource memory endpoints, so the scans can touch exactly those resources that fit the requested memory range without having to iterate over every single allocated resource.

This gives me around a 3x speedup in Xenoblade after the intro scene.


exjam:
Looks good, we need to get around to cleaning up old resources at some point though!

PR #389, opened 2016-12-31
Interpreter fixes.

achurch:
Various things I found while working on libbinrec.

FYI, I've also expanded the test suite at http://achurch.org/cpu-tests/ppc750cl.s in case you want to rebuild the associated rpx (there are I think 4 tests which still fail, mostly from lfd+ps_* data corruption).

### DIFF
--- a/src/common/bitutils.h
+++ b/src/common/bitutils.h
@@ -184,6 +184,21 @@ clz(uint32_t bits)
 #define clz __builtin_clz
 #endif
 
+#ifdef PLATFORM_WINDOWS
+inline int
+clz64(uint64_t bits)
+{
+   unsigned long a;
+   if (!_BitScanReverse64(&a, bits)) {
+      return 64;
+   } else {
+      return 63 - a;
+   }
+}
+#else
+#define clz64 __builtin_clzll
+#endif
+
 inline bool
 bit_scan_reverse(unsigned long *out_position, uint32_t bits)
 {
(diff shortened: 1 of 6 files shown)


exjam:
For "espresso: Fix reversed operand order in rlwimi disassembly."

Could we change the src/libcpu/espresso/espresso_instruction_definitions.inl instead?

from:
INS(rlwimi, (rA), (**rS, rA**, sh, mb, me), (rc), (opcd == 20), "Rotate Left Word Immediate then Mask Insert")

to:
INS(rlwimi, (rA), (**rA, rS**, sh, mb, me), (rc), (opcd == 20), "Rotate Left Word Immediate then Mask Insert")

The rest of the changes look good.

I assume all of your interpreter accuracy based changes you found whilst developing your new JIT? I took a look at it, it certainly looks great and I have no objections to integrating it as the JIT engine in decaf. Hopefully this month I can find some time to see if I can help debug the issue you see with the GPU thread running significantly slower.

achurch:
Fixed, thanks for catching that. (The operand naming for the logic instructions always confuses me; I assume they did it to simplify circuitry in the CPU, but it's still weird that they use a different field for the output register than almost every other instruction.)

I did figure out the problem with the GPU thread, btw -- it seems like the NVIDIA Linux driver is really bad at waiting on queries, because when I added an explicit wait loop on GL_QUERY_RESULT_AVAILABLE before fetching the query result, the slowness went away (see e064b5c7). I have no clue why that should be the case, but I don't feel like debugging a closed-source driver and the change shouldn't have any significant performance impact relative to a properly-working driver, so I'm just going to leave it at that for now.

I'll go ahead and put up a PR for the libbinrec change; feel free to ping me if you have any questions. There are more optimizations I'm planning to add, but I've got a new work project starting up so I probably won't have much time for libbinrec in the near future.

PR #390, opened 2016-12-31
Reimplement GPU synchronization.

achurch:
The previous implementation had huge overhead (I measured a frame rate drop of close to 50% in Xenoblade) and wrapped several operations which arguably don't need wrapping, such as simple data writes.

### DIFF
--- a/src/libdecaf/src/gpu/opengl/opengl_driver.cpp
+++ b/src/libdecaf/src/gpu/opengl/opengl_driver.cpp
@@ -409,6 +409,26 @@ GLDriver::getGpuClock()
    return coreinit::OSGetTime();
 }
 
+static uint64_t
+swapValueForWrite(uint64_t value, latte::CB_ENDIAN swap)
+{
+   switch (swap)
+   {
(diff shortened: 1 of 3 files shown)


exjam:
@brett19  ?

brett19:
This is really a performance vs. correctness situation.  A simple example of a case where games would behave badly due to the simplified synchronization would be a case of doing a feedback operation, and then inserting a data-write in the stream to detect when the feedback operation completed (or any other operation like this, for instance, geometry counting).  In this case, because the data-write is not synchronized but the feedback operation is, the data-write would occur immediately, giving the game the impression that the data-write has completed (since its expecting serial execution), when in fact it has not.

achurch:
I'm pretty sure I accounted for cases like that, but honestly it's been so long since I made this change that I don't recall the details of my thought processes. If you can give me an actual (not theoretical) example of a game that breaks because of this change, I'll take a second look, but otherwise, this is what I have to offer and it's yours to do with as you like.

PR #391, opened 2017-01-03
New JIT implementation using libbinrec.

achurch:


Margen67:
Does this perform better than the old implementation?

CreeperMario:
> Does this perform better than the old implementation?
JIT is faster than interpretation in Dolphin, so I'm going to assume yes.

achurch:
>Does this perform better than the old implementation?

It depends on the specific patterns of code in the game, but with all optimizations enabled (--jit-opt-level=3 --jit-fast-math=full), in Xenoblade I measured an average of about 50% reduction in CPU time compared to the old JIT implementation. The old implementation fell back to the interpreter for a number of instructions (notably paired-single load and store), so code which uses a lot of those instructions will see markedly better performance.

It's also worth noting that libbinrec defaults to accurate (interpreter-equivalent) behavior if no optimizations are enabled, which makes floating-point code slower than the previous JIT implementation because of the time required to update FPSCR correctly.

achurch:
The AppVeyor build has been broken for a long time. I don't use Windows, so someone else will need to fix that.

exjam:
We need to move the Windows build to using CMake in order to prevent these problems in the future!

jroweboy:
@exjam also need to make the appveyor use a appveyor.yml instead of just using appveyor settings so we can work around libbinrec being a mercurial repo

(Also i should dust off my old cmake on windows changes :p)

naters305:
now that exjam has converted to CMake for windows. Will the appveyor build of this issue pass? I feel that this JIT improvement could help the current performance of decaf quite well.

achurch:
I've never tried building libbinrec with the Visual Studio compiler so I don't know whether any issues will crop up there (feel free to file bug reports on Bitbucket if they do). I'll also see about getting this change rebased, though I may not get to it for a bit because of work.

exjam:
I'm gonna play around with it today, this is my notes so far:

Added dependencies: mercurial, git-remote-hg
To get the submodule need to add GIT_ALLOW_PROTOCOL=hg
GIT_ALLOW_PROTOCOL=hg git submodule update --init


To build libbinrec for Windows there are 2 changes required:

1. min/max is defined by stdlib on Windows, so I suggest a ifndef min/max

```
1>E:\wiiu\projects\decaf-emu\libraries\libbinrec\src/common.h(266): warning C4005: 'min': macro redefinition [E:\wiiu\projects\decaf-emu\build\libraries\libbinrec-prefix\src\libbinrec-build\binrec.vcxproj]
1>    C:\Program Files (x86)\Windows Kits\10\Include\10.0.14393.0\ucrt\stdlib.h(1290): note: see previous definition of 'min'
```

```
//stdlib.h:1290
    #ifndef __cplusplus
        #define max(a,b) (((a) > (b)) ? (a) : (b))
        #define min(a,b) (((a) < (b)) ? (a) : (b))
    #endif
```

2. For MSVC it is __restrict instead of restrict, so I suggest you add a define in common.h to handle this difference between the compilers:
```
2>E:\wiiu\projects\decaf-emu\libraries\libbinrec\src\host-x86\host-x86-translate.c(49): error C2146: syntax error: missing ';' before identifier 'buffer' [E:\wiiu\projects\decaf-emu\build\libraries\libbinrec-prefix\src\libbinrec-build\binrec.vcxproj]
```

exjam:
I believe currently libbinrec is violating Windows x64 calling convention
https://msdn.microsoft.com/en-us/library/ms235286.aspx 

> The x64 Application Binary Interface (ABI) uses a four register fast-call calling convention by default. Space is allocated on the call stack as a shadow store for callees to save those registers. 

So I guess this means you must allocate 8*4 = 32 bytes at least (+8 for mxscr) on the stack for any function which calls another function, @brett19  will know for sure.

You can see this in the following snippets, the r11 function is jit.cpp:195 scHandler(Core *core), and the first thing it does is overwrite rsp+8 which is the location we used for stmxcsr and ldmxcsr in the calling function (because call puts 8 bytes on stack for return address)

```
0000000200000050  sub         rsp,8  
0000000200000054  mov         r11d,10010E4h  
000000020000005A  mov         dword ptr [rcx+4],r11d  
000000020000005E  mov         r11,qword ptr [rcx+358h]  
0000000200000065  stmxcsr     dword ptr [rsp]  
0000000200000069  call        r11
000000020000006C  ldmxcsr     dword ptr [rsp]  
0000000200000070  add         rsp,8 
```

CALL r11 ->
```
--- e:\wiiu\projects\decaf-emu\src\libcpu\src\jit\jit.cpp ----------------------
00007FF77F97ABF0  mov         qword ptr [rsp+8],rcx 
...
```

achurch:
Thanks for the notes -- I've pushed fixes to the repo.

>min/max redefined

Fixed by just undefining them first.

>restrict vs. __restrict

Fixed (though restrict is required by C99 so Microsoft ought to fix their compiler).

>!!int

As a matter of style, I prefer to avoid the double exclamation mark and to omit explicit "!=0" when zero indicates failure / absence of a value (see: http://achurch.org/SIL/current/STYLE.html#comparing-against-zero). I recommend disabling that warning if it bothers you.

>Argument shadow space

I thought I had added this, but apparently not -- thanks for catching it. Fixed.

exjam:
I am seeing some bizzarre behaviour with MSVC, looks like #if IS_MSVC(1, 0) does not actually work.

http://rextester.com/AAI8951

```
#if (defined(_MSC_VER) && _MSC_VER >= (1 * 100 + 0))
# pragma message("IS_MSVC 1") // WORKS
#else
# error "NOT_MSVC 1"
#endif

#define IS_MSVC(major, minor) (defined(_MSC_VER) && _MSC_VER >= (major * 100 + minor))

#if IS_MSVC(1,0)
# error "IS_MSVC 2"
#else
# error "NOT_MSVC 2" // FAILS
#endif
```

Gamedev1909:
so ive compiled this and how do i get it to run what like --jit or more to it

achurch:
--jit --jit-opt-level=2 is safe but slow.
--jit --jit-opt-level=3 --jit-fast-math=full is fast but may cause some games to crash or not work correctly.

exjam:
So I just tested it and it seems to be completely ignoring a branch instruction.

The code highlighted in IDA is the basic block that your JIT has compiled:
https://puu.sh/usY1s/ac1e192dba.png

Code from your JIT:
https://puu.sh/usY0b/a17d9555ee.png

I can see it loads the r3 value, (the `mov r11d,dword ptr [rcx+14h]`) then performs the compare, but then it just seems to do nothing for the beq instruction? And consequently goes through to the `call r11` at the end which is the call to jit::trapHandler (aka the tw instruction in ppc)

achurch:
Ooh, that's interesting. And very odd because it's at the top of a basic block, so there shouldn't be anything known about r3 which would trigger elision of the branch.

I'm having a bit of trouble trying to reproduce this here, even if I explicitly delete the branch. What optimization flags did you enable?

exjam:
opt_flags: ["BASIC"] is enough to eliminate the branch entirely, check this gist for the PowerPC and x64 output for no opt_flags and opt_flags = BASIC

https://gist.github.com/exjam/d08122f71b50de2ecd8d2208923307e8

achurch:
Doesn't happen for me with only BASIC enabled: (full output: https://gist.github.com/achurch/32089d13801cad599c27419ee272bc94 -- the relevant PPC code is splatted into a random convenient place in memory so some addresses won't match)
```
   0x00007fffddf39e60:  and    $0x20000000,%r8d
   0x00007fffddf39e67:  test   %r8d,%r8d
   0x00007fffddf39e6a:  je     0x7fffddf39e8b
```
This is on Linux, but there's nothing in the PPC->RTL translator or RTL core that looks at the host platform, so that shouldn't make a difference. I've also run the libbinrec tests through valgrind, so I'm pretty confident there's no uninitialized data that could trigger this.

Can you dump the RTL from before and after finalization, just to be certain where the branch is getting dropped? I don't have an API for this at the moment, but here's a quick-and-dirty line you can insert into src/api.c:
```
extern int printf(const char*,...); extern void fflush(void*); extern void *stdout; printf("%s\n",rtl_disassemble_unit(unit,1)); fflush(stdout);
```
Useful places to dump would be after the rtl_finalize_unit() call (around line 575) and after the rtl_optimize_unit() call immediately following it.

Also perhaps try disabling MSVC optimizations when compiling libbinrec? Wouldn't be the first compiler bug I've run into...

(FYI, I've updated this commit and libbinrec with separate fixes for invalid instructions, but they should be unrelated to this problem.)

exjam:
finalised_0x%X.txt is right after rtl_finalize_unit
optimised_0x%X.txt is right after rtl_optimize_unit

https://gist.github.com/exjam/150ac4762e8adb91be5ec2c65a45476c

I have a branch on here which has the changes I made to get decaf and libbinrec running on Windows btw, so I'm not working off this PR commit if you updated it with any actual code change.
https://github.com/decaf-emu/decaf-emu/tree/binrec

achurch:
Okay, I think I finally see what's going on. Are you building with ENABLE_ASSERT turned off? In that case I turn assert() calls into __assume on MSVC (see ASSERT macro in common.h), but it looks like MSVC doesn't actually execute code within the __assume expression. I've committed a change to libbinrec to remove that special case and use the generic version; let me know if that fixes the problem.

exjam:
I'm not convinced that is the issue - I commented out the call to rtl_opt_drop_dead_blocks(unit); and the branch remains (that doesn't mean I know that the final code is correct, but it has the branch at least!).

If RTL_DEBUG_OPTIMIZE helps, this is the output (this is _without_ me commenting out the call):
```
[35:29.513340 debug:w108] [libbinrec] Address: 03ABBAA0
[35:29.517807 debug:w108] [libbinrec] Dropping dead block 16 (155-156)
[35:29.517935 debug:w108] [libbinrec] Killing instruction 156
[35:29.518009 debug:w108] [libbinrec] r1 death rolled back to 153
[35:29.518056 debug:w108] [libbinrec] r96 no longer used, setting death = birth
[35:29.518098 debug:w108] [libbinrec] Killing instruction 155
[35:29.518135 debug:w108] [libbinrec] Dropping dead block 14 (147-148)
[35:29.518170 debug:w108] [libbinrec] Killing instruction 148
[35:29.518205 debug:w108] [libbinrec] r93 no longer used, setting death = birth
[35:29.518239 debug:w108] [libbinrec] Killing instruction 147
[35:29.518274 debug:w108] [libbinrec] Dropping dead block 9 (108-109)
[35:29.518310 debug:w108] [libbinrec] Killing instruction 109
[35:29.518378 debug:w108] [libbinrec] r69 no longer used, setting death = birth
[35:29.518418 debug:w108] [libbinrec] Killing instruction 108
[35:29.518455 debug:w108] [libbinrec] Dropping dead block 4 (44-45)
[35:29.518489 debug:w108] [libbinrec] Killing instruction 45
[35:29.518535 debug:w108] [libbinrec] r30 no longer used, setting death = birth
[35:29.518577 debug:w108] [libbinrec] Killing instruction 44
[35:29.518613 debug:w108] [libbinrec] Dropping branch at 154 to next insn
[35:29.518647 debug:w108] [libbinrec] Killing instruction 154
```

When I comment out the call there is no killed instructions, branches or blocks.

If it helps, this is the full PPC instruction: https://gist.github.com/exjam/c0a6d530555909bab72330a2267307a8

achurch:
Did you check to see whether the branch in question was present? That's at instruction 18, which isn't touched by any of the RTL_DEBUG_OPTIMIZE output you pasted. There are always going to be some dead instructions/blocks/branches because of how the RTL is generated.

The problem was that __assume was deleting the function call which added fallthrough edges from one block to the next -- if you look at the RTL dumps, you can see from the pre-optimzation dump (https://gist.github.com/exjam/150ac4762e8adb91be5ec2c65a45476c#file-finalised_0x3abbaa0-txt) that most blocks have no incoming edges and no block has an edge to its successor. The optimizer detects a dead block as one which has no incoming edges and is not the entry block, so (among others) it drops blocks 1 and 2, which immediately follow the branch at 18. The drop_dead_branches pass then sees that the branch at 18 is a branch to the next instruction, so it gets killed as well.

achurch:
@exjam FYI, the changes currently missing on your binrec branch are here: https://github.com/achurch/decaf-emu/commit/9db73ca7006a64720f562a120682572ffb234463 (plus a libbinrec update - I've added a load forwarding optimization that resolves the problem of ASSUME_NO_SNAN corrupting data copied with lfs/stfs).

exjam:
Just checked this again, it's still hitting a trap at the start of every game (because every single game seems to do a branch around a tw instruction). (Using the latest libbinrec, Fri Mar 24 21:50:36 2017 +0900).

This includes your favourite game Xenoblade! So this must be a Windows specific issue if you managed to test Xenoblade on Linux succesfully...?

Maybe I should refactor libcpu to better support multiple backends so we can keep both around until this one is stable on Windows. Someone on IRC has requested this refactor before anyway.

achurch:
That's very odd. Is there any chance you could get the libbinrec tests building on Windows? (It should just be a matter of compiling each test, linking with libbinrec, and running, but you'll want to automate it because there are about 3300 of them.) The tests all pass on Linux, so maybe that could help pinpoint the problem.

exjam:
So from understanding the Makefile, I just have to build a single .c file for each test (+ linking it to libbinrec) which match the filters:

tests/*/*.c
tests/*/*/*.c
tests/api/binrec++.cc

And then running them is as simple as checking the return code?

Sounds simple enough, I'll add that to the CMakeLists.txt and see what happens. I have to do actual work right now but I will play with the tests tonight.

achurch:
That looks right, yes. All test programs exit with EXIT_SUCCESS (0) on success and EXIT_FAILURE (looks like it might be -1 in MSVC? Should always be nonzero anyway) on failure. For some tests, you can set LOG_VERBOSE=1 in the environment to dump log messages to stdout (useful with failing tests).

exjam:
Things which don't compile on msvc2017:

This initialisation syntax is not supported:
tests/guest-ppc/\*/\*.c: `#define INITIAL_STATE  &(PPCInsnTestState){.gqr = {0,0,0x00040000}}`

This initialisation syntax is not supported:
tests/execute.c: `(DWORD[1]){}`

Empty arrays are not supported:
tests/host-x86/\*/\*.c `static const uint8_t expected_code[] = {};`

Also my poor hard drive...! Running CMake generates ~23k files just for the build files...!! Then when you compile a test you get like 10 more files per test just for the intermediate files!

I did a partial test run from some which compiled before I started hitting too many compile errors:
https://gist.github.com/exjam/d3a4f321b3d7a56f4d9d83ad2e4d66dd/raw/be01baf42c16a439c6894a43104feb746151f0ac/binrec%2520partial%2520windows%2520run

There are a lot of failures..! Something is very wrong with binrec on Windows! Not sure if the tests which managed to run help, I gave up fixing compile errors when i hit the INITIAL_STATE define about ~10% of the way into the test compiling.

exjam:
Oh I forgot to run with LOG_VERBOSE.
Done that now (same small 10% of the tests) and it's 900k+ lines long and 18mb .txt file so I uploaded it to google drive: https://drive.google.com/file/d/0B9yE_xHd2F_6c1B2b0pheWtUV0E/view?usp=sharing

achurch:
I've committed a fix for the empty arrays, but the other two syntaxes are valid C99, so I guess those are MSVC bugs. I'll see if I can come up with some sort of workaround.

One quick note about your test output -- it looks like you have RTL_DEBUG_OPTIMIZE defined for the library build but not the tests, so a bunch of tests are reporting spurious failures because of the log differences. That said, the failures of tests/guest-ppc/exec/750cl-* suggest there is indeed a problem somewhere. I scanned the output and didn't see anything obvious, but the spurious failures make it a bit harder to spot any problems.

>Also my poor hard drive...!

I admit I've gotten used to having an SSD. (: That said, I assume there are probably compiler switches to reduce the number of auxiliary files created? You probably don't need things like debug symbol files, for example.

achurch:
Try this patch for the C99 compound literals: http://achurch.org/libbinrec-msvc-hack.diff

exjam:
Oh it still doesnt like
`static PPCInsnTestState _initial_state = {.gqr = {}};`

Its fixes it by doing
`static PPCInsnTestState _initial_state = {.gqr = {0}};`

Assuming those two statements are equivalent?

achurch:
Yeah, that's fine.

achurch:
On further investigation, apparently an empty initializer list is not in fact permitted by C99. Which seems strange, but live by the standard, die by the standard, I suppose. I'll commit a fix for that.

exjam:
Also, for msvc at least, adding the {0} works for the dummy_void / dummy_long stuff aswell, we can use `(DWORD[1]){0}` whilst `(DWORD[1]){}` was a compiler error.

exjam:
The sra / srl tests are failing with an ASSERT.

For example with sra-bmi2-reuse-src1.exe:
 append_vex_opcode
 ASSERT(opcode > 0xFF);
 opcode = X86OP_SARX (0xf30f38f7)

Looks like signed opcode comparison to >0xFF is failing? I guess you need to force X86Opcode to be 64bit, or to be unsigned, or to change the assert to cast to unsigned first.

exjam:
Oh and guest-ppc\exec\750cl-chain.exe crashes completely, not sure why: https://puu.sh/v1hPT/298effeded.png

Here's an (almost) complete test run logs, I fixed the missing flags so less tests fail, compile failed at 99% because I forgot to define VERSION, but this is 99% of the tests! Didn't want to wait another 30mins for the tests to compile again just to get the last 1%....!

https://drive.google.com/file/d/0B9yE_xHd2F_6VU1OWnF3ck1XUXM/view?usp=sharing

achurch:
I could have sworn that constants >= 0x80000000 were always unsigned (if 32-bit), but I'll look at the C99 reference again... Does adding a "u" suffix to the enumerator values fix the asserts?

I'll take a look at the test logs now.

achurch:
I don't actually see any other significant failures in the test log; the FP failures under rtl/* suggest that MSVC isn't properly implementing fma(), but that shouldn't have any effect on integer code. Also, the native_features test suggested that you're using a CPU without BMI2, so the SARX issue shouldn't have come up while executing code (which makes sense - if it had, the code stream would have fallen apart at the point of such an instruction anyway).

I'm a bit puzzled by all the 750cl-* failures at the moment, but I'll keep looking.

I do see that C99 requires all enum values to fit in a signed int, so I'll fix that now. Live and learn...

achurch:
On further consideration, it looks like all the 750cl-* failures (except that one crash, which I suspect has something to do with how I'm allocating memory for the generated code in the tester and is not an actual failure) can be explained by the generated code not using FMA3 instructions, because those tests are all to check for non-fused operation (which is what you get if FMA3 isn't available).

If you link that same libbinrec library (.lib / .dll / however you built it) into decaf-emu, do you still get the trap failure?

exjam:
OK so it's Debug mode vs RelWithDebInfo mode is the problem here. I was building the tests with Debug, and with decaf-emu i was building RelWithDebInfo.

So now when the tests are built in RelWithDebInfo mode every test crashes...

```
Test project E:/wiiu/projects/libbinrec2-build
        Start   1: guest-ppc_exec_750cl-assume-no-snan-fast-stfs
  1/3296 Test   #1: guest-ppc_exec_750cl-assume-no-snan-fast-stfs .................***Exception: SegFault  1.71 sec

        Start   2: guest-ppc_exec_750cl-assume-no-snan
  2/3296 Test   #2: guest-ppc_exec_750cl-assume-no-snan ...........................***Exception: SegFault  0.97 sec

        Start   3: guest-ppc_exec_750cl-branch-exit-test
  3/3296 Test   #3: guest-ppc_exec_750cl-branch-exit-test .........................***Exception: SegFault  1.32 sec

        Start   4: guest-ppc_exec_750cl-chain
  4/3296 Test   #4: guest-ppc_exec_750cl-chain ....................................***Exception: SegFault  1.15 sec

        Start   5: guest-ppc_exec_750cl-constant-gqrs
  5/3296 Test   #5: guest-ppc_exec_750cl-constant-gqrs ............................***Exception: SegFault  5.17 sec

        Start   6: guest-ppc_exec_750cl-fast-fmadds
  6/3296 Test   #6: guest-ppc_exec_750cl-fast-fmadds ..............................***Exception: SegFault  1.30 sec

        Start   7: guest-ppc_exec_750cl-fast-fmuls
  7/3296 Test   #7: guest-ppc_exec_750cl-fast-fmuls ...............................***Exception: SegFault  0.82 sec

        Start   8: guest-ppc_exec_750cl-fast-stfs
  8/3296 Test   #8: guest-ppc_exec_750cl-fast-stfs ................................***Exception: SegFault  0.77 sec

        Start   9: guest-ppc_exec_750cl-fnmadd-zero-sign
  9/3296 Test   #9: guest-ppc_exec_750cl-fnmadd-zero-sign .........................***Exception: SegFault 37.57 sec

        Start  10: guest-ppc_exec_750cl-forward-loads
 10/3296 Test  #10: guest-ppc_exec_750cl-forward-loads ............................***Exception: SegFault  1.78 sec

        Start  11: guest-ppc_exec_750cl-hostopt
 11/3296 Test  #11: guest-ppc_exec_750cl-hostopt ..................................***Exception: SegFault  0.81 sec

        Start  12: guest-ppc_exec_750cl-ignore-fpscr-vxfoo
.... etc ...
```

exjam:
Oh right its because we need BINREC_ENABLE_ASSERT = FALSE on Windows.

achurch:
Enabling/disabling assertions shouldn't make a difference - everything asserted should be true anyway (and I fixed the bug with MSVC ignoring the expression). If there are crashes, that's probably a real bug.

(also shoutouts to GitHub for sleeping on the job and not showing new comments)

exjam:
Well if you have BINREC_ENABLE_ASSERT = TRUE then you are doing `#include <assert.h>` which under MSVC is a no-op on none debug builds:

```
#ifdef NDEBUG
    #define assert(expression) ((void)0)
```

So on MSVC you have to set BINREC_ENABLE_ASSERT  = FALSE to disable binrec from including and using the Windows assert.h which does nothing on release builds.

achurch:
Oh wow, that's fascinating (and probably a violation of the standard). Fair enough.

exjam:
Got it running on Windows now 👍  https://puu.sh/v284T/088487b2de.jpg

Just going to spend some time cleaning up it's integration with decaf.

achurch:
FWIW, it turns out NDEBUG is part of C99 too, so I've committed a fix for that to libbinrec. I guess that comes from the same dark corners of ancient history as things like the getopt() global variables...

exjam:
Hey achurch, I've almost finished this PR off! There's still a few changes I want to do, such as fixing breakpoints for the imgui debugger. I was thinking of making it fallback to interpreter for a code block which has breakpoints, although I guess that wouldn't work if the enable chaining optimisation is on. Although if we wanted to add breakpoint support for when enable chaining is on, maybe we could do an old the fashioned JMP hook to interpreter land at start of whatever function had a BP in it? Is it safe to do a JMP hook from the start of a JIT entry point to a C++ function which calls into interpreter?

We thought it might be best to host a git mirror of the hg repository on bitbucket to avoid adding a dependency on hg + hg-git. Is that okay with you? It would be a read only mirror of your bitbucket.

Mirror:
https://github.com/decaf-emu/libbinrec

My current binrec branch:
https://github.com/decaf-emu/decaf-emu/tree/binrec https://github.com/decaf-emu/decaf-emu/commit/4c2049bf47eca2fd40b6955b026f262beed22511

achurch:
The git mirror is fine with me.

I don't think there's any good way to use breakpoints with chaining -- as soon as you have to return to an intermediate handler, it's probably just as fast to not chain at all. Though I suppose you could generate a custom trampoline for each chain lookup that does the breakpoint check and then jumps to the known target address, which would at least save the lookup cost each time through.

exjam:
Well I think it might be ready to go now, I implemented breakpoints by setting the PPC instruction to a tw instruction and invalidating the code cache for any code block which could hit that breakpoint (start -> start +4096).

When the JIT hits a tw, it returns to the execution loop where it will fall back to the interpreter where the breakpoint can be handled gracefully. In the interpreter when it executes a tw instruction it will lookup the current address in the breakpoint list to find the original saved PPC code and execute that instead.

Tested: add breakpoint / remove breakpoint, break_on_entry, step over, step into.

I will run my latest changes on Linux quickly to test then I'll push it to master, scary!

exjam:
Merged as 4c2049bf47eca2fd40b6955b026f262beed22511

PR #399, opened 2017-03-23
multiprocessor compilation switch for msvc compilers

takhlaq:
(ﾉ◕ヮ◕)ﾉ*:･ﾟ✧

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,7 +22,9 @@ endif()
 
 if(MSVC)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /std:c++latest")
-
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MP")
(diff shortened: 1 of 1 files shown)


VelocityRa:
Wanted that but didn't want to touch CMAKE, thanks :D

CMakeLists.txt
```diff
@@ -22,7 +22,9 @@ endif()
 
 if(MSVC)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /std:c++latest")
```
jroweboy: minor nit pick, but you can throw them all on the same line.

set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /std:c++latest /MP /FS")

takhlaq: yeah, decided to not one line it in case anyone needs to remove it for whatever reason and so it's clearer to see separate flags

PR #400, opened 2017-03-24
kernel: Take .thrbss into account when calculating TLS size.

achurch:
For a module with a .thrbss section but no .thrdata, the kernel would incorrectly calculate a TLS size of zero.

### DIFF
--- a/src/libdecaf/src/kernel/kernel_loader.cpp
+++ b/src/libdecaf/src/kernel/kernel_loader.cpp
@@ -1042,13 +1042,28 @@ loadRPL(const std::string &moduleName,
 
    // Calculate TLS base
    auto thrdata = findSection(sections, shStrTab, ".thrdata");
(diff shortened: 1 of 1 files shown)


exjam:
Do you think we should make sure that we can handle if .thrbss vaddr comes before .thrdata vaddr too? Or do you think it's a safe assumption to make that thrbss will always come after thrdata.

exjam:
Saying that, I only have two games which have both a .thrdata and .thrbss, WUP-P-BH8E and WUP-P-AUME, and for both thrbss comes after thrdata.

achurch:
I think the only way you could get .thrbss before .thrdata is if somebody deliberately modifies linker scripts and crt0/etc to put BSS before initialized data, and I can't imagine any sensible reason for doing that. I'll add in an assertion just in case, but it's probably safe to assume .thrbss follows .thrdata.

PR #404, opened 2017-03-27
Fix install step for glbinding

XenonPK:
On linux, ${CMAKE_INSTALL_CONFIG_NAME} is not part of the glbinding builddir, which causes it to complain when running make install, as the file is not in the expected folder.


exjam:
We made glbinding static on Linux, this should fix the install issue!
https://github.com/decaf-emu/decaf-emu/commit/077105336ddc4b561a8db93a2c734935dd38ee55

PR #410, opened 2017-04-04
jit: Inline code block lookups in the execution loop.

achurch:
This reduces overall CPU usage by around 2% in my tests.


exjam:
Seems sensible, I'm sure there's many of these minor optimisation changes that can be made!

PR #413, opened 2017-04-04
jit: Support new libbinrec optimization flags.

achurch:
No huge benefit from these, but a percent here, a percent there...

This will need a libbinrec update, naturally.


exjam:
LGTM, btw I added you on the decaf-emu/libbinrec repo so you are able to push to it when you make updates like if you wish - if not I don't mind maintaining the mirror as it's just a `git pull origin && git push decaf-mirror` for me.

I think we will need to sort out the config shit out soon - this duplication between decaf-cli and decaf-sdl is silly.

Also I'm 100% 💯  💯  👌  up for moving away from cereal for the config file as it's just too strict with the config file format....!

PR #416, opened 2017-04-04
Make glbinding optional.

achurch:
While glbinding is useful in ensuring correctness of OpenGL code, it also imposes significant runtime overhead; I have observed slowdown of up to 35% (compared to direct OpenGL calls) in GPU-bound code, with the GPU thread spending over 15% of its time just in malloc() and free() calls from glbinding.


exjam:
I believe as long as we do **not** call glbinding::setAfterCallback then glbinding will almost directly map OpenGL functions pointers without anything fancy surrounding it.

Could you try profile with deleting the setCallbackMaskExcept and setAfterCallback function calls here:
https://github.com/decaf-emu/decaf-emu/blob/20eb74bfacf7ddf3bd79583f471f88dbe76871c0/src/decaf-sdl/decafsdl_opengl.cpp#L132
https://github.com/decaf-emu/decaf-emu/blob/cf2bb40a363155b9fc362b807778292e780a49e7/tools/pm4-replay/sdl_window_draw.cpp#L106

Well I guess you could just include those in the `if (decaf::config::gpu::debug) {` statement right below it.

I believe this would achieve the same functionality than this PR, you can see what glbinding does here:
https://github.com/cginternals/glbinding/blob/master/source/glbinding/include/glbinding/Function.inl#L153-L172

I mean there is still a minor overhead of checking the isAnyEnabled state, but it's probably small enough to go unnoticed to even a profiler.

achurch:
Unfortunately, that doesn't seem to help:

(Current codebase unchanged - second field is % time consumed)
```
16997     5.6223  libc-2.23.so             _int_free
15777     5.2188  libc-2.23.so             malloc
12503     4.1358  libc-2.23.so             _int_malloc
1848      0.6113  libc-2.23.so             free
571       0.1889  libc-2.23.so             malloc_consolidate
```

(With setAfterCallback disabled)
```
23300     5.9434  libc-2.23.so             _int_free
20906     5.3327  libc-2.23.so             malloc
16909     4.3132  libc-2.23.so             _int_malloc
2616      0.6673  libc-2.23.so             free
696       0.1775  libc-2.23.so             malloc_consolidate
```

(With glbinding removed - %s are slightly higher, but that's probably just random variance)
```
703       0.4834  libc-2.23.so             _int_malloc
543       0.3733  libc-2.23.so             _int_free
515       0.3541  libc-2.23.so             malloc
120       0.0825  libc-2.23.so             malloc_consolidate
111       0.0763  libc-2.23.so             free
```

glbinding::FunctionCall shows up a bunch in the malloc/free call graphs, so I imagine the various levels of templates in glbinding end up needing dynamic allocation somewhere.

achurch:
Correction - looks like I only disabled setAfterCallback and not setCallbackMaskExcept. With both of those disabled, the overhead is gone:
```
597       0.4320  libc-2.23.so             _int_malloc
486       0.3517  libc-2.23.so             malloc
467       0.3380  libc-2.23.so             _int_free
161       0.1165  libc-2.23.so             malloc_consolidate
118       0.0854  libc-2.23.so             free
```

I'll create a separate PR for that. Though I'll probably keep this on my local branch just so I don't have to build glbinding every time I clean the build tree...

exjam:
Great! Sorry that you lost time on that!

PR #418, opened 2017-04-04
Build glbinding as a static instead of shared library.

achurch:
This allows running the program from the build directory without needing to manually set LD_LIBRARY_PATH on Linux.


exjam:
Fixed with https://github.com/decaf-emu/decaf-emu/commit/077105336ddc4b561a8db93a2c734935dd38ee55

PR #420, opened 2017-04-05
swkbd: Fix GetInputFormString() returning wrong byte order under GCC.

achurch:
I can't find an explicit reference in the C++11 spec, but it appears that the encoding of wchar_t is implementation-defined (such that ```(int)L'a' == 'a'``` doesn't necessarily hold), while char16_t is simply "a 16-bit-wide char". At any rate, the cppreference.com example for UTF8-to-UTF16 conversion uses char16_t, so hopefully this formulation should be safe everywhere.

### DIFF
--- a/src/libdecaf/src/modules/swkbd/swkbd_core.cpp
+++ b/src/libdecaf/src/modules/swkbd/swkbd_core.cpp
@@ -125,7 +125,7 @@ GetDrawStringInfo(DrawStringInfo *info)
 const be_val<uint16_t> *
 GetInputFormString()
 {
-   std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>> converter;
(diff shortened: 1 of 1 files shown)


exjam:
I notice your change fails on MSVC, seems like for some reason it doesn't like a codecvt with char16_t and instead wants int16_t.

Would this change work OK?:
https://github.com/exjam/decaf-emu/commit/1129e6d428d238f0420dc229b457ff01e6b910a1

achurch:
Yup, seems to work fine for me.

exjam:
1129e6d428d238f0420dc229b457ff01e6b910a1

exjam:
p.s. LLE swkbd.rpl almost works... https://puu.sh/v9X6W/2b5615b142.png

Will have to implement some stuff in nsyskbd.rpl to fix the bugs though with that though!

PR #424, opened 2017-04-13
Fix munmap of incorrect address.

petmac:
The previous code checked if the mapped address (result) matches the desired address (baseAddress). If it does not match, the previous code attempted to unmap the desired address rather than the mapped address.

This fix unmaps the mapped address.

### DIFF
--- a/src/common/src/platform_posix_memory.cpp
+++ b/src/common/src/platform_posix_memory.cpp
@@ -35,7 +35,7 @@ reserveMemory(size_t address, size_t size)
 
    if (result != baseAddress) {
       if (result != nullptr) {
-         freeMemory(address, size);
+         munmap(result, size);
       }
 
       return false;


src/common/src/platform_posix_memory.cpp
```diff
@@ -35,7 +35,7 @@ reserveMemory(size_t address, size_t size)
 
    if (result != baseAddress) {
       if (result != nullptr) {
-         freeMemory(address, size);
+         munmap(result, size);
```
brett19: I suspect this needs to use freeMemory rather than munmap.

brett19:
I think you used the wrong function to unmap the memory.

petmac:
Hi,

Thanks for checking this PR so quickly. That's what it used to call, but I favoured munmap for 2 reasons:
- freeMemory just casts the address back to a pointer and calls munmap, so I skipped 2 casts.
- Semantically, the memory isn't reserved until reserveMemory returns, so it's the mmap that should be undone, rather than the reserveMemory.

Happy to change it, the end result will be the same.

brett19:
Woops.  I should have looked at this a bit more, looks like my initial implementation had more than one weird problem 👍

petmac:
Thanks!

Issue #426, opened 2017-04-15
Potential macOS version discussion


petmac:
Hi,

I'm interested in doing some or all of the work required to port decaf-emu to macOS.

I've done some work on it independently, and the emulator builds and runs.

The current roadblock is that macOS only supports OpenGL 4.1 rather than the current target of 4.5. This doesn't stop the emulator running (after some version number changes) but it does prevent anything from being drawn. Some functions and some GLSL features are missing.

In order to not waste anyone's time, I wanted to discuss the following scenarios:

1. A macOS version is not wanted, in which case I'll just stop working on it.
2. A macOS version is wanted, but it's not acceptable to downgrade the OpenGL requirement on Windows or Linux to 4.1. Assuming macOS can support everything decaf-emu needs, a separate macOS-compatible renderer (i.e., Metal or OpenGL 4.1) would be required.
3. A macOS version is wanted, and it's acceptable to downgrade to OpenGL 4.1.

Thanks,
Peter

brett19:
Hey @petmac,

@exjam and I have had this discussion on numerous occasions, as we both actually use MBP's for work.  The main problem with downgrading OpenGL to 4.1 is that it legitimately does not support the features that are required to render the WiiU games, it also has some performance implications.

We had previously considered implementing a Metal rendering engine, similar to how we've implemented our DX12 rendering engine, but the interop between Obj-C and C++11 is non-trivial (although just searching now, I found this new project which looks promising: https://github.com/naleksiev/mtlpp).  I have however read a few things as well about Obj-C <-> C++ interop being difficult to get leak free due to differing semantics.

All of this being said, we would be ecstatic to see a community member contribute a Metal rendering backend.  I'd be more than happy to provide any necessary support towards the goal of implementing a Metal backend.

Cheers, Brett

exjam:
Yeah it basically comes to Apple really dropping the ball on it's graphics driver situation. Even Metal isn't sufficient as it lacks features such as Geometry Shaders. And looking at https://en.wikipedia.org/wiki/OpenGL#OpenGL_4.1 looks like there has been some useful OpenGL stuff between 4.1 and 4.5.

I mean you can try - but it's going to be rather challenging to emulate the missing features.

Gamedev1909:
Well use metal to vulkan wrapper its free called moltelvk works just like vulkan

exjam:
MoltenVK does not support geometry shaders.

Gamedev1909:
[...]

PR #427, opened 2017-04-16
Build for macOS.

petmac:
This pull request makes enough changes to allow the code to build in Xcode for macOS.

As mentioned in #426, we cannot use OpenGL 4.1 because it doesn't have required features. The plan then is to implement a Metal rendering engine.

This pull request does not make a start on that, but it does disable the use of OpenGL on macOS. I didn't make changes to use a stubbed/null renderer, so the emulator won't run anything, but it's enough to build the code.

I will implement a stubbed Metal renderer in a future pull request.

It would be useful for me if someone could check that this build runs OK on Linux or Windows. I don't run either OS so apart from Travis succeeding, I didn't verify I hadn't broken anything.


src/common/src/platform_posix_exception.cpp
```diff
@@ -70,8 +70,12 @@ dispatchException(Exception *exception,
          return;
       } else {
          // Exception handled, switch execution to target function
-         auto ctx = reinterpret_cast<ucontext *>(context);
+         auto ctx = reinterpret_cast<ucontext_t *>(context);
+#ifdef __APPLE__
```
exjam: Instead of using __APPLE__ in code directly, can you use PLATFORM_APPLE?

https://github.com/decaf-emu/decaf-emu/blob/master/src/common/platform.h#L6

PR #430, opened 2017-04-24
posix: Use TMPDIR instead of shm_open for memory backing files.

achurch:
The amount of memory available for shm objects is limited on some
systems, so use a normal file instead of an shm object as the backing
store for emulated memory.

Also document the cause and workaround for SIGBUS crashes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,3 +49,5 @@ It is recommend to run the emulator from the root git directory so that it is ab
 Configuration files can be found at:
 - Windows - `%APPDATA%\decaf`
 - Linux - `~/.config/decaf`
+
+On Linux, a "Bus error" crash usually indicates an out-of-space error in the temporary directory.  Set the `TMPDIR` environment variable to a directory on a filesystem with at least 2GB free.
--- a/src/common/src/platform_posix_memory.cpp
+++ b/src/common/src/platform_posix_memory.cpp
@@ -41,17 +41,28 @@ getSystemPageSize()
 MapFileHandle
 createMemoryMappedFile(size_t size)
 {
-   int fd = shm_open("/decaf", O_RDWR | O_CREAT | O_EXCL, 0600);
+   const char *tmpdir = getenv("TMPDIR");
+   if (!tmpdir || !*tmpdir) {
+      tmpdir = "/tmp";
+   }
+   const std::string pattern = fmt::format("{}/decafXXXXXX", tmpdir);
+   char *path = strdup(pattern.c_str());  // Must be a modifiable char array.
+   int old_umask = umask(0077);
+   int fd = mkstemp(path);
    if (fd == -1) {
-      gLog->error("createMemoryMappedFile({}) shm_open failed with error: {}",
+      gLog->error("createMemoryMappedFile({}) mkstemp failed with error: {}",
                   size, errno);
+      umask(old_umask);
       return InvalidMapFileHandle;
    }
+   umask(old_umask);
 
-   if (shm_unlink("/decaf") == -1) {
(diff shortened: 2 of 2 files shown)


achurch:
Ninja edit - forgot to check for a present-but-empty $TMPDIR value...

exjam:
Is it better to use shm_open on systems where we can use it? I was concerned that by using a file for the memory mapping then Linux might have some obligation to actually write the data to a file and thus not perform as well as using shm_open.

linkmauve:
On a modern Linux system, you can use `memfd_create(2)` instead, so you only have to do the `shm_open(2)`/`unlink(2)` dance in the fallback case on systems where memfd isn’t supported.

achurch:
>I was concerned that by using a file for the memory mapping then Linux might have some obligation to actually write the data to a file

For Linux, at least, shm_open is essentially a wrapper around open() on a special in-memory filesystem (and the problem is that that filesystem may be mounted with a size limit -- e.g. 64MB on my system), so there's no difference versus opening a file on a memory-backed (tmpfs) filesystem. In any case, blocks are flushed asynchronously, so there's no performance hit on actual stores to the mapped region. If you feel strongly about keeping shm_open(), I could add a config option for it, but there needs to be at least the possibility of using a regular file in case the shm filesystem gets full.

>On a modern Linux system, you should use memfd_create(2) instead

That's non-POSIX. The POSIX version works just as well with tmpfs backing storage.

linkmauve:
Better remount your /dev/shm with a proper size really, there is no reason to have such a small tmpfs for something that important on your system.

And yes, memfd is specific to Linux, that’s exactly what I said.

achurch:
>Better remount your /dev/shm with a proper size really, there is no reason to have such a small tmpfs for something that important on your system.

Funny how no other software ever runs into problems with that size. I'll manage my system in a way that works for me, thanks.

>And yes, memfd is specific to Linux, that’s exactly what I said.

In which case there's no reason to use it, since we need POSIX code for Mac anyway.

exjam:
So after some discussion on IRC, maybe we should revisit this using facebook's hhvm as inspiration:
https://github.com/facebook/hhvm/blob/1a9516a957e8763216ee24c0eaa2bfdbd93809cc/hphp/hack/src/heap/hh_shared.c#L570-L642

https://github.com/facebook/hhvm/blob/1a9516a957e8763216ee24c0eaa2bfdbd93809cc/hphp/hack/src/heap/hh_shared.c#L138-L177

It seems they try to use mem_fd on systems where available, with falling back to shm_open on Apple or to mkstemp on non-Apple.

It seems they use shm_dir as a configuration value which defaults to /dev/shm but also seems like they have a way to specify multiple shmdirs with one as TMPDIR:
https://github.com/facebook/hhvm/blob/10b4a1aeb38979d8632ab7496ddab390777f6b2d/hphp/hack/src/globals/globalConfig.ml#L29-L51

achurch:
>memfd_create seems ideal because due to not having a mount point it will not be limited by any file size limitations

On the flip side, it is limited by available memory (physical + swap) and it's forced to be stored in memory, whereas having a backing file gives the OS an option to swap it out elsewhere - sort of like a private swap area. Granted, that's probably not a huge consideration for the typical environment in which Decaf would be used (particularly since I think Ubuntu defaults to putting /tmp on tmpfs anyway), but it could relieve memory pressure while Decaf is in the background.

I still don't think it's worth the extra effort to have Linux-specific code, but (as long as TMPDIR is also an option) I'm not going to object if someone else writes it. (:

PR #435, opened 2017-05-01
Debugger Performance Window (fps graphs & graphics info)

VelocityRa:
Currently looks like this:

image

There's probably lots of debugging info that could be added here.

#### Feedback/suggestions welcome!

### DIFF
--- a/src/libdecaf/decaf_graphics_info.h
+++ b/src/libdecaf/decaf_graphics_info.h
@@ -1,0 +1,20 @@
+#pragma once
+
+#include <cstdint>
+
+namespace decaf
+{
+
+struct GraphicsDebugInfoGL {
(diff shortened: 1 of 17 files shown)


exjam:
Maybe we should be making the "graphics debugging info" backend specific rather than using a shared interface for all back ends. For instance, there's some stuff which would apply to Vulkan / Metal but not OpenGL (e.g. the manual memory management, pipeline state objects, etc). I like the idea though.

We have an abstraction for OpenGL graphics driver specific functions: https://github.com/decaf-emu/decaf-emu/blob/master/src/libgpu/gpu_opengldriver.h

Those graphs are sexy tho 👀👌💯🔥

exjam:
Merged this with refactorings.

PR #442, opened 2017-06-05
Modified shader related files.

dogkarl:
Modifed related files to allow MaxTextures to be 18.
Also modified the shader translation to allow SAMPLER instruction to use this much texture.
Added SAMPLER_LB implementation.

Trying to run Zelda breath of wild :-)

### DIFF
--- a/src/libgpu/latte/latte_enum_sq.h
+++ b/src/libgpu/latte/latte_enum_sq.h
@@ -247,6 +247,7 @@ ENUM_BEG(SQ_SEL, uint32_t)
    ENUM_VALUE(SEL_W,                      3)
    ENUM_VALUE(SEL_0,                      4)
    ENUM_VALUE(SEL_1,                      5)
+   ENUM_VALUE(SEL_BIAS,                      6)
    ENUM_VALUE(SEL_MASK,                   7)
 ENUM_END(SQ_SEL)
 
--- a/src/libgpu/latte/latte_enum_sq.h
+++ b/src/libgpu/latte/latte_enum_sq.h
@@ -247,7 +247,7 @@ ENUM_BEG(SQ_SEL, uint32_t)
    ENUM_VALUE(SEL_W,                      3)
    ENUM_VALUE(SEL_0,                      4)
    ENUM_VALUE(SEL_1,                      5)
-   ENUM_VALUE(SEL_BIAS,                      6)
+   ENUM_VALUE(SEL_BIAS,                   6)
    ENUM_VALUE(SEL_MASK,                   7)
 ENUM_END(SQ_SEL)
 
--- a/src/libgpu/src/glsl2/glsl2_tex.cpp
+++ b/src/libgpu/src/glsl2/glsl2_tex.cpp
@@ -145,12 +145,18 @@ sampleFunc(State &state,
 
    auto resourceID = inst.word0.RESOURCE_ID();
    auto samplerID = inst.word2.SAMPLER_ID();
+   auto lodBias = inst.word1.LOD_BIAS();
 
    auto samplerDim = state.shader->samplerDim[samplerID];
-   auto samplerUsage = registerSamplerID(state, samplerID, isShadowOp);
+   //Use resourceID to set the Usage
+   auto samplerUsage = registerSamplerID(state, resourceID, isShadowOp);
 
-   if (resourceID != samplerID) {
-      throw translate_exception("Unsupported sample with RESOURCE_ID != SAMPLER_ID");
+   //The statement 'resourceID == samplerID' is true when the texture number is below 16.
+   //However, while sampler ID is limited to 16 max,max texture number allowed is 18. 
+   //So when use texture at index 16 and texture at index 17,the sampler id will be 0 and 1.
+   //When compile shader, the compiler will warn about this sampler overlap.
+   if (resourceID%16 != samplerID) {
+      throw translate_exception("Unsupported sample with RESOURCE_ID%16 != SAMPLER_ID");
    }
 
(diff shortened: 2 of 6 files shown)


src/libgpu/latte/latte_enum_sq.h
```diff
@@ -247,6 +247,7 @@ ENUM_BEG(SQ_SEL, uint32_t)
    ENUM_VALUE(SEL_W,                      3)
    ENUM_VALUE(SEL_0,                      4)
    ENUM_VALUE(SEL_1,                      5)
+   ENUM_VALUE(SEL_BIAS,                      6)
```
VelocityRa: Spacing

src/libgpu/src/glsl2/glsl2_tex.cpp
```diff
@@ -145,12 +145,18 @@ sampleFunc(State &state,
...
-      throw translate_exception("Unsupported sample with RESOURCE_ID != SAMPLER_ID");
+   //Use resourceID to set the Usage
+   auto samplerUsage = registerSamplerID(state, resourceID, isShadowOp);
+
+   //The statement 'resourceID == samplerID' is true when the texture number is below 16.
+   //However, while sampler ID is limited to 16 max,max texture number allowed is 18. 
+   //So when use texture at index 16 and texture at index 17,the sampler id will be 0 and 1.
+   //When compile shader, the compiler will warn about this sampler overlap.
+   if (resourceID%16 != samplerID) {
+      throw translate_exception("Unsupported sample with RESOURCE_ID%16 != SAMPLER_ID");
```
VelocityRa: * `sample` -> `sampler` or perhaps
* change the message to something like "TextureID > 16. Sampler overlap will occur."

src/libgpu/src/glsl2/glsl2_tex.cpp
```diff
@@ -181,8 +187,9 @@ sampleFunc(State &state,
          decaf_check(func.size());
          state.out << func;
       }
-
-      state.out << "(sampler_" << samplerID << ", ";
+	  
+	  //Here we should not use samplerID
```
VelocityRa: * Maybe you should explain you don't use the samplerID here as well.
* Indent with 1 more space
* A space after the `//` after every one of your comments wouldn't hurt either.

src/libgpu/src/glsl2/glsl2_translate.cpp
```diff
@@ -586,9 +586,9 @@ insertFileHeader(State &state)
       }
    }
 
-   // Samplers
-   for (auto id = 0u; id < state.shader->samplerDim.size(); ++id) {
-      auto dim = state.shader->samplerDim[id];
+   // Samplers,use the number of samplerUsage for cycle
+   for (auto id = 0u; id < state.shader->samplerUsage.size(); ++id) {
+      auto dim = state.shader->samplerDim[id%state.shader->samplerDim.size()];
```
VelocityRa: * Could extract `state.shader->samplerDim.size()` to a new `const` variable.
* Might want to space things out when using the module operator, like with all operators.

SakataGintokiYT:
@dogkarl 
Now i got this error 

```
File: C:\projects\decaf-emu\src\libgpu\src\opengl\opengl_texture.cpp
Line: 159
Message: Unimplemented texture xy filter 5
Stacktrace:
15:  - 0x0x
14:  - 0x0x
13:  - 0x0x
12:  - 0x0x
11:  - 0x0x
10:  - 0x0x
9:  - 0x0x
8:  - 0x0x
7:  - 0x0x
6:  - 0x0x
5:  - 0x0x
4:  - 0x0x
3:  - 0x0x
2: crt_at_quick_exit - 0x7FFB197CBDA0x
1: BaseThreadInitThunk - 0x7FFB1B6D80E0x
0: RtlUserThreadStart - 0x7FFB1DFFC580x
```

src/libgpu/src/glsl2/glsl2_translate.h
```diff
@@ -69,7 +69,9 @@ struct Shader
    std::string codeBody;
    std::vector<Export> exports;
    std::array<std::vector<Feedback>, latte::MaxStreamOutBuffers> feedbacks;
-   std::array<SamplerUsage, latte::MaxSamplers> samplerUsage;
+   //The Usage array should be length of MaxTexture.Because the last 2 texture
```
VelocityRa: * Space after `//`
* `texture` -> `textures`

dogkarl:
TO SakataGintokiYT 
I've only fixed the shader error. Now I am facing the same error too, I am still trying to understand the part of code that deals with textures. My guess is now the max texture number is 18, the part the manages the actual texture also need fixing.

src/libgpu/latte/latte_enum_sq.h
dogkarl: Should I fix these style problems and then recommit? I am quite new at GitHub stuffs, so many thanks for any advise and guidance.
And Thanks for all your detailed advice.

VelocityRa: Yes, that is the easiest option if you're new to git.
They can all be squashed down to 1 commit later, so no problem there.
You're welcome :)

exjam:
I think with max textures being 18 and max samplers being 16 you highlight the more important problem that GX2 does not require SamplerID to match TextureID - you can use the same sampler with different textures. However with OpenGL you can only use a single texture with a sampler, which is why the assert of SamplerId == ResourceId is there.

The correct fix is to analyse the shader and for every Sampler x Texture combination we must generate an OpenGL/GLSL sampler. This is a big pain the ass which is why I have not done it yet. In an ideal world we would use Vulkan with Spir-V where we can properly emulate GX2 with separate sampler and texture objects...

Issue #456, opened 2017-09-10
Buggy/inconsistent JIT implementation?


Kinnay:
When issue #455 is fixed, DKC:TF will show its splash screen. If JIT is disabled, it will even get to the title screen. However, with JIT enabled the game usually crashes somewhere during the loading screen. "Usually", because sometimes decaf itself crashes before even showing the loading screen. Here's a screenshot from the debugger:

![image](https://user-images.githubusercontent.com/10852324/30252807-2ad5f07a-9679-11e7-943d-1e4b9ef59411.png)

I also ran the JIT verifier a few times. This is what usually happened (even the jit verifier doesn't always behave the same):

```[06:45.744404 critical:w108] Assertion failed:
Expression: core->gpr[i] == coreCopy->gpr[i]
C:\Users\Yannik\Desktop\files\WiiU\decaf-emu\decaf-emu\src\libcpu\src\jit\binrec\jit_binrec_verify.cpp
Line: 529
Message: Wrong value in GPR 11 at 0x224492C: lwz r11, r3, 8
      Found: 0x00000000
   Expected: 0x00000001
Stacktrace:
4: platform::captureStackTrace - 0x7FF7739E9C60x
3: assertFailed - 0x7FF7739E7FC0x
2: cpu::jit::BinrecBackend::verifyPost - 0x7FF773DC5660x
1: cpu::jit::BinrecBackend::brVerifyPostHandler - 0x7FF773DC4BD0x
0: cpu::jit::BinrecBackend::brVerifyPostHandler - 0x7FF773DC4BD0x```

Kinnay:
I may have found the solution to the first problem (decaf crashing with an access violation).

The problem lies here: https://github.com/decaf-emu/decaf-emu/blob/10eeebd1d6dac26037474f4ef4291c5658291eb2/src/libcpu/src/jit/binrec/jit_binrec.cpp#L283-L297

On Windows, the buffer pointer is increased by codeOffset, and the correct codeSize is calculated by the next line. However, the original buffer size is passed to mCodeCache.registerBlock. Because the size of the buffer is now smaller than its original size, an out-of-bounds read occurs when CodeCache::registerBlock tries to copy the buffer. If the memory behind the buffer is not readable, decaf-emu crashes.

I think this can be fixed by simply passing codeSize to mCodeCache.registerBlock instead of the "size" variable.

exjam:
Yeah I believe your analysis is correct, codeSize should work there. PR? :)

[...]

PR #461, opened 2017-09-14
Fix a crash in jit_binrec.cpp

Kinnay:
Issue: https://github.com/decaf-emu/decaf-emu/issues/456


xaviex:
Should the issue for this be closed now that this is fixed?

Kinnay:
There's still a bug that's making the JIT compiler behave differently than the interpreter. That's what I'm referring to in the part about the JIT verifier. So while the first part of the issue has been fixed, the other part is still an issue.

PR #464, opened 2017-09-24
Fix the Windows binary link

jlachniet:
Previously, this link was completely incorrect. Should be fixed now.


exjam:
Can you change this to add ?branch=master at the end - otherwise it will show builds from pull requests or other branches.

exjam:
Thanks!

PR #465, opened 2017-09-26
libdecaf: If cos.xml is missing, default to the first .rpx we find

VelocityRa:
Are multiple .rpx's even an issue?

### DIFF
--- a/src/libdecaf/src/decaf.cpp
+++ b/src/libdecaf/src/decaf.cpp
@@ -1,3 +1,4 @@
+#include "common/platform_winapi_string.h"
 #include "decaf.h"
 #include "decaf_config.h"
 #include "decaf_graphics.h"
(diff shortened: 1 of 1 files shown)


exjam:
I've been avoiding using std::filesystem yet because not all compilers seem to support it in a sane way yet. Includes are either <filesystem> or <experimental/filesystem>. Namespace can be seen as std::filesystem, std::experimental::filesystem, or std::experimental::filesystem::v1.

I mean strictly speaking we could add a bunch of defines and try figure out the right combo per compiler, but I'd rather hold off using std::filesystem until it is ready. Once it is we can rewrite a bunch of platform specific code we currently have.

If you want to push through with this API then I suggest modifying common/platform_dir.h to add filesystem functions which perform the equivalent behaviour per platform (Windows + posix).

VelocityRa:
Ah I see.

>If you want to push through with this PR then I suggest modifying common/platform_dir.h to add filesystem functions which perform the equivalent behaviour per platform (Windows + posix).

It's not really an important feature, so I don't think it's worth it if the fs platform-specific code is going to be rewritten anyway.

PR #466, opened 2017-10-08
Fix for C++17 code on macOS

CreeperMario:
The changes in this PR are quite minimal.

* Travis will build Decaf using the Xcode 9 image instead of the Xcode 8.3 image.
* Xcode 9.0 (the latest official compiler for macOS) does not yet fully support C++17, and so macOS builds have been configured to use `-std=c++1z` instead of `-std=c++17`. My assumption is that Xcode 9.1 will be based on LLVM v5, which contains completed support for C++17, at which point, this specific change can be reversed.
* CMake will error if the user attempts to build on macOS with `-DDECAF_GL=ON`
* In several places throughout `libdecaf`, I have casted some of the values sent to fmtlib to either `be_val<uint8_t>` or `be_val<uint16_t>`. `clang` seems to be having trouble automatically handling template expansion when passing certain enumerations to `fmtlib` for error printing, but manually casting them seems to suppress the error. The values that were casted were already `uint8_t` or `uint16_t` under the hood, so this should not cause any issues. These changes do not affect the emulator itself, this only changes the behaviour when Decaf spits out error messages.

At this point, Travis should have no trouble delivering builds for macOS.


CreeperMario:
So unfortunately, the macOS build is still failing, but for reasons completely unrelated to Decaf.

It seems to have had trouble setting up the Homebrew package manager, spitting out an error related to Ruby version mismatches. At this point, I believe that this is an error on Travis' side, as other users seem to be suffering the same issue.

I'll be following travis-ci/travis-ci#8552 and will update as necessary.

exjam:
Thanks for trying to sort this out, I did try myself when I initially updated to C++17 but couldn't figure it out. This reminds me I that I forgot about your wut PR, i'll go sort that out now :).

CreeperMario:
😄

exjam:
I merged this manually - removed the fmt related changes because they were due to a bug in fmt lib which has now been fixed https://github.com/fmtlib/fmt/issues/580

Issue #449, opened 2017-06-29
Zelda BOTW - fastMutexQueue.head


xsellier:
Hi folks,

Keep up the good work !

Another tiny issue since the last push

```
Assertion failed:
Expression: !thread->fastMutexQueue.head
File: /opt/decaf-emu/src/libdecaf/src/modules/coreinit/coreinit_fastmutex.cpp
Line: 376

Assertion failed:
Expression: thread->state == OSThreadState::Running
File: /opt/decaf-emu/src/libdecaf/src/modules/coreinit/coreinit_thread.cpp
Line: 1195

Assertion failed:
Expression: thread->state == OSThreadState::Running
File: /opt/decaf-emu/src/libdecaf/src/modules/coreinit/coreinit_thread.cpp
Line: 1195
```

PR #469, opened 2017-12-26
Implement unlockAllFastMutexNoLock

schulmar:
A first try at implementing this function.
I am not sure if this achieves what the inline comment intended but it solves the deadlock/assertions (e.g. #449)  for me.

### DIFF
--- a/src/libdecaf/src/modules/coreinit/coreinit_fastmutex.cpp
+++ b/src/libdecaf/src/modules/coreinit/coreinit_fastmutex.cpp
@@ -194,7 +194,7 @@ fastMutexHardUnlock(OSFastMutex *mutex)
 }
 
 static void
-OSFastMutex_Unlock_common(OSFastMutex *mutex) {
+OSFastMutex_Unlock_one_NoLock(OSFastMutex *mutex) {
     decaf_check(mutex);
     decaf_check(mutex->tag == OSFastMutex::Tag);
     decaf_check(mutex->count > 0);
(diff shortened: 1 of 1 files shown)


src/libdecaf/src/modules/coreinit/coreinit_fastmutex.cpp
```diff
@@ -193,46 +193,59 @@ fastMutexHardUnlock(OSFastMutex *mutex)
    internal::unlockScheduler();
 }
 
+static void
+OSFastMutex_Unlock_common(OSFastMutex *mutex) {
```
brett19: This should probably be NoLock postfixed?

src/libdecaf/src/modules/coreinit/coreinit_fastmutex.cpp
```diff
@@ -193,46 +193,59 @@ fastMutexHardUnlock(OSFastMutex *mutex)
...
+          continue;
+       }
+
+       // Success!
+       break;
+    }
+}
+
+void
+OSFastMutex_UnlockWithoutScheduler(OSFastMutex *mutex) {
```
brett19: NoScheduler is probably more in line withe our existing naming standards.

src/libdecaf/src/modules/coreinit/coreinit_fastmutex.cpp
brett19: Additionally, OSFastMutex_Unlock_one might be more in line with our standards, and makes it more clear what it does.

schulmar: Well, you tell me ;) I don't know about your naming conventions and was not sure whether NoLock was directed at the mutex itself or the scheduler.

brett19:
@exjam Can you review/merge?

exjam:
Sorry for the slow response - I have been away.

Unfortunately your fix is not correct. NoLock implies the function is already holding the scheduler lock - and unfortunately your code would lead to the scheduler possibly being incorrectly unlocked due to unlockAllFastMutexNoLock -> OSFastMutex_UnlockWithoutScheduler -> OSFastMutex_Unlock_common -> fastMutexHardUnlock -> unlockScheduler. Also unlockAllFastMutexNoLock is not meant to lead to  testThreadCancel or reschedule called, which could happen if fastMutexHardUnlock is hit.

So basically the main problem here is that unlockAllFastMutexNoLock is actually a distinct code path from OSFastMutex_Unlock. As such it is better to just implement it in place:
https://github.com/decaf-emu/decaf-emu/commit/2c51cd4eedc77ff67fa34ef5223ec805985c8e6e

Whilst I was there I also found 2 bugs in usage of isContended:
https://github.com/decaf-emu/decaf-emu/commit/c27e76db820ce8879fadb603ca1f70f36fbafbb4

Thanks for bringing this to my attention, I was unaware that this function had been left unimplemented (I blame Brett!). For future note, reversing coreinit.rpl is a great reference for implementation details such as these. I hope this doesn't discourage you from attempting more contributions in the future :).

PR #472, opened 2018-01-07
latte-assembler: fixes/improvements.

aliaspider:


tools/latte-assembler/src/gfd.cpp
```diff
@@ -325,7 +493,7 @@ gfdAddPixelShader(gfd::GFDFile &file,
 
    out.regs.sq_pgm_resources_ps = out.regs.sq_pgm_resources_ps
       .NUM_GPRS(numGpr)
-      .STACK_SIZE(1);
+      .STACK_SIZE(out.loopVars.size() * 2);
```
aliaspider: I'm still not entirely sure about this one. there is definitely some correlation between loopVarsCount and stacksize though.

brett19:
I see Geometry Shader related stuff here, there may be useful information here as well:
https://github.com/decaf-emu/decaf-emu/tree/geom

aliaspider:
the assembler doesn't support geometry shaders yet, those changes you saw were only for libgfd/gfd-tool.

exjam:
Merged, thanks.

PR #473, opened 2018-02-01
Compile fix for pm4-replay-qt

cornytrace:
Hi there, I was checking the project out of interest, and found if DECAF_QT was enabled, pm4-replay-qt would fail to compile (on windows). I believe I resolved the issue but I just started looking at the code, so please check if my fix is correct.


exjam:
Looks good thanks - I don't often build the Qt code so this one slipped past me!

PR #474, opened 2018-02-02
Remove DX12

cornytrace:
Was only a stub anyway. Vulkan is a better, cross-platform solution.


shadowbrony33:
I second this, Vulkan is better.

brett19:
Hey Guys,
Just to provide some insight here, the reason that we originally worked to implement DX12 as opposed to Vulkan is directly related to the fact DX12's debugging and error handling capabilities far out-weighted the benefit of having an additional cross-platform backend.  OpenGL already provides the ability for us to make cross-platform builds of Decaf and implementing an entirely new framework for rendering (direct control of the GPU) required a significant amount of trial and error, where good analysis and debugging tools are highly desirable.  In addition to this, the DX12 and Vulkan APIs are very similar in their semantics, so we believe it would be possible to initially develop the DX12 version, using all its tools, and then convert that to a Vulkan backend in the future rather trivially.
Cheers, Brett

exjam:
Also Vulkan wasn't even released when we were doing dx12. The very first graphics backend for decaf was actually dx12, then when we rewrote how graphics backends work I went with OpenGL instead. We also have a branch where the dx12 backend was being brought back since the rewrite: https://github.com/decaf-emu/decaf-emu/tree/dx12

I mainly wrote these stubs because Brett was working on the dx12 backend the same time I did a bunch of huge refactorings so I wanted to make sure it was not too difficult for him to port his dx12 backend into my refactored code.

Regardless of the history I do not see any dx12 development going on anytime soon so I guess we can remove these stubs. I would like to start work on a Vulkan graphics backend but I do have other things I need to finish first, maybe I'll get started on it like H2 2018 :D.

PR #489, opened 2018-09-13
cafe/snd_user: Implement FxGetMemSize 

SteveLeafo:
Started an implementation of snd_user which Progresses New Super Mario Bros. U. I am hoping add some missing sounds to Gianna Sisters so more to come.

Not sure what happened to the last pull request - had a stray end of line character.

### DIFF
--- a/src/libdecaf/src/ios/acp/ios_acp.cpp
+++ b/src/libdecaf/src/ios/acp/ios_acp.cpp
@@ -1,5 +1,7 @@
 #include "ios_acp.h"
+#include "ios_acp_acp_main_thread.h"
 #include "ios_acp_client_save.h"
+
 #include "ios/fs/ios_fs_fsa_ipc.h"
 #include "ios/kernel/ios_kernel_heap.h"
 #include "ios/kernel/ios_kernel_messagequeue.h"
@@ -116,6 +118,7 @@ processEntryPoint(phys_ptr<void> /* context */)
 
    // Initialise static memory
    internal::initialiseStaticData();
+   internal::initialiseStaticAcpMainThreadData();
 
    auto error = IOS_SetThreadPriority(CurrentThread, 50);
    if (error < Error::OK) {
@@ -141,6 +144,12 @@ processEntryPoint(phys_ptr<void> /* context */)
    messageQueueId = static_cast<MessageQueueId>(error);
 
    error = MCP_RegisterResourceManager("/dev/acpproc", messageQueueId);
+   if (error < Error::OK) {
+      return error;
+   }
+
+   // Start thread for /dev/acp_main
+   error = internal::startAcpMainThread();
    if (error < Error::OK) {
       return error;
    }
--- a/src/libdecaf/src/ios/acp/ios_acp_acp_main_thread.cpp
+++ b/src/libdecaf/src/ios/acp/ios_acp_acp_main_thread.cpp
@@ -1,0 +1,100 @@
+#include "ios_acp_acp_main_thread.h"
+
+#include "ios/ios_handlemanager.h"
+#include "ios/ios_stackobject.h"
+#include "ios/kernel/ios_kernel_process.h"
+#include "ios/kernel/ios_kernel_resourcemanager.h"
+#include "ios/kernel/ios_kernel_thread.h"
+#include "ios/mcp/ios_mcp_ipc.h"
+
+namespace ios::acp::internal
+{
+
+using namespace kernel;
+
+constexpr auto AcpMainNumMessages = 100u;
+constexpr auto AcpMainThreadStackSize = 0x10000u;
+constexpr auto AcpMainThreadPriority = 50u;
+
+struct StaticAcpMainThreadData
+{
+   be2_val<ThreadId> threadId;
+   be2_val<MessageQueueId> messageQueueId;
+   be2_array<Message, AcpMainNumMessages> messageBuffer;
+   be2_array<uint8_t, AcpMainThreadStackSize> threadStack;
+};
+
+static phys_ptr<StaticAcpMainThreadData>
+sAcpMainThreadData = nullptr;
+
+static Error
+acpMainThreadEntry(phys_ptr<void> /*context*/)
+{
+   StackObject<Message> message;
+
+   while (true) {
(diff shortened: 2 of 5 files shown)


tgsm:
some advice for rebasing on git:
* i really recommend creating your own branch for your changes rather than trying to merge from your `master` branch.
* if at any point you need to update your fork to be more update w/ decaf:
1. `git fetch upstream` (i'm assuming you know how to add remotes)
2. `git rebase upstream/master`
3. `git push -f origin <your branch name>`

src/libdecaf/src/cafe/libraries/snd_user/snd_user_fx.h
```diff
@@ -0,0 +1,99 @@
...
+   virt_ptr<int32_t> surround;
+};
+CHECK_OFFSET(AXFXBuffers, 0x00, left);
+CHECK_OFFSET(AXFXBuffers, 0x04, right);
+CHECK_OFFSET(AXFXBuffers, 0x08, surround);
+CHECK_SIZE(AXFXBuffers, 0x0C);
+
+struct AXFXReverbHi
+{
+   float       *earlyLine[3];
```
exjam: `be2_array<virt_ptr<float>, 3> earlyLine;`

src/libdecaf/src/cafe/libraries/snd_user/snd_user_fx.h
```diff
@@ -0,0 +1,99 @@
...
+   uint32_t     earlyPos[3];
```
exjam: `be2_array<uint32_t, 3> earlyPos`

src/libdecaf/src/cafe/libraries/snd_user/snd_user_fx.h
```diff
@@ -0,0 +1,99 @@
...
+   uint32_t     earlyLength;
```
exjam: be2_val<uint32_t> earlyLength;

src/libdecaf/src/cafe/libraries/snd_user/snd_user_fx.h
```diff
@@ -0,0 +1,99 @@
...
+   float        crosstalk;        // crosstalk of each channels
+   float        earlyGain;        // output gain of early reflection (0.f - 1.f)
+   float        fusedGain;        // output gain of fused reverb (0.f - 1.f)
+
+   AXFXBuffers *busIn;
+   AXFXBuffers *busOut;
+   float        outGain;
+   float        sendGain;
+};
+
```
exjam:
```
CHECK_OFFSET(AXFXReverbHi, 0x0, earlyLine)
...
CHECK_SIZE(AXFXReverbHi, xxx)
```

src/libdecaf/src/cafe/libraries/snd_user/snd_user_fx.h
```diff
@@ -0,0 +1,99 @@
...
+
+struct AXFXDelay
```
exjam: Same for this structure.

src/libdecaf/src/cafe/libraries/snd_user/snd_user_fx.h
```diff
@@ -0,0 +1,99 @@
...
+      virt_ptr<void>(uint32_t size)
+>;
+
+using AXFXFreeFuncPtr = virt_func_ptr<
+void(virt_ptr<void> ptr)
+>;
+
+struct AXFXBuffers
+{
+   virt_ptr<int32_t> left;
```
exjam: be2_virt_ptr

src/libdecaf/src/cafe/libraries/snd_user/snd_user_fx.cpp
```diff
@@ -0,0 +1,43 @@
+#include "snd_user.h"
+#include "snd_user_fx.h"
+#include "cafe/libraries/cafe_hle_stub.h"
+
+namespace cafe::snd_user
+{
+
+static AXFXAllocFuncPtr
```
exjam: One day we will support process switching so we should not use host level static variables, but instead store them in the rpl .data section. The RegisterDataInternal allows us to swap the host pointer when the userland process changes, which is actually what the call to `coreinit::internal::relocateHleLibrary` is doing in the rpl_entry - one day we will also call that on process switch.

See how we do static data in other rpls:
https://github.com/decaf-emu/decaf-emu/blob/master/src/libdecaf/src/cafe/libraries/coreinit/coreinit_alarm.cpp#L23-L41

```
struct StaticFxData
{
be2_val<AXFXAllocFuncPtr> allocFuncPtr;
be2_val<AXFXFreeFuncPtr> freeFuncPtr;
};

static virt_ptr<StaticFxData>
sFxData = nullptr;
```

https://github.com/decaf-emu/decaf-emu/blob/master/src/libdecaf/src/cafe/libraries/coreinit/coreinit_alarm.cpp#L576
`RegisterDataInternal(sFxData);`

exjam:
Oh boy multidimensional arrays do not work nicely with be2_array
` be2_array<virt_ptr<be2_array<virt_ptr<float>, 2>>, 3> `

I'll see if I can come up with improvement to be2_array to make this better...

SteveLeafo:
Yeah - thanks, that looks a lot better and the offsets make more sense now too.

exjam:
Are you sure you got these correct - how did you figure all the structures out? I am generally dubious of structures all filled out with no code using them because there is no way to know they're correct.

For example if we look at the AXFXDelayGetMemSize function from snduser2.rpl:
https://puu.sh/BvQRB/0eaca3ae2b.png
https://puu.sh/BvQNs/dec580dcbf.png

You can see it adds up 3 values in an array at 0x40 into the AXFXDelay structure, then multiplies by 4 then by (AXGetInputSamplesPerSec() / 1000). This implies to me at 0x40 there is an array of 3 values of the MS delay amount for this effect, probably like a per audio source or something I dunno I have not looked into AXFX much yet.

But in your structures at 0x40 there is "active", at 0x44 delay buffer length per channel which sounds like the correct thing but what is a "channel" here? In sndcore2 we currently mix to 2 channels (L / R) but sndcore2.rpl actually supports up to 6 channel audio.

PR #490, opened 2018-09-16
More implementation of snduser2_axfx

SteveLeafo:
Sorry guys - this is the same changes as before including feedback, but (hopefully) without the merge issues


exjam:
Are you sure you got these correct - how did you figure all the structures out? I am generally dubious of structures all filled out with no code using them because there is no way to know they're correct.

For example if we look at the AXFXDelayGetMemSize function from snduser2.rpl:
https://puu.sh/BvQRB/0eaca3ae2b.png
https://puu.sh/BvQNs/dec580dcbf.png

You can see it adds up 3 values in an array at 0x40 into the AXFXDelay structure, then multiplies by 4 then by (AXGetInputSamplesPerSec() / 1000). This implies to me at 0x40 there is an array of 3 values of the MS delay amount for this effect, probably like a per audio source or something I dunno I have not looked into AXFX much yet.

But in your structures at 0x40 there is "active", at 0x44 delay buffer length per channel which sounds like the correct thing but what is a "channel" here? In sndcore2 we currently mix to 2 channels (L / R) but sndcore2.rpl actually supports up to 6 channel audio.

exjam:
Thanks, merged as eb4f353104f163b6941fa7798c9c05d2e52dbfb9

PR #491, opened 2018-09-26
Update README.md

niansa:
Made the cmake command more simple...


VelocityRa:
The command is like that to allow for out-of-source builds.

baka0815:
Most times you do ``mkdir build;cd build;cmake ..`` or something similar to nit build into the source tree.

exjam:
Builds inside the source tree are discouraged as a bad practice.

PR #495, opened 2018-10-05
Implementation for H.264 decoding

SteveLeafo:
Sorry about the last few pull requests - still a little bit new to Git - but I think I have it all worked out now.

### DIFF
--- a/src/libdecaf/src/cafe/libraries/h264/bs.h
+++ b/src/libdecaf/src/cafe/libraries/h264/bs.h
@@ -1,0 +1,388 @@
+/* 
+ * h264bitstream - a library for reading and writing H.264 video
+ * Copyright (C) 2005-2007 Auroras Entertainment, LLC
+ * Copyright (C) 2008-2011 Avail-TVN
+ * 
+ * Written by Alex Izvorski <aizvorski@gmail.com> and Alex Giladi <alex.giladi@gmail.com>
+ * 
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ * 
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ * 
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+#ifndef _H264_BS_H
+#define _H264_BS_H        1
+
+#include <stdint.h>
+#include <stdlib.h>
+#include <string.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef struct
+{
+	uint8_t* start;
+	uint8_t* p;
+	uint8_t* end;
+	int bits_left;
+} bs_t;
+
+#define _OPTIMIZE_BS_ 1
+
+#if ( _OPTIMIZE_BS_ > 0 )
+#ifndef FAST_U8
+#define FAST_U8
+#endif
+#endif
+
+
+static bs_t* bs_new(uint8_t* buf, size_t size);
+static void bs_free(bs_t* b);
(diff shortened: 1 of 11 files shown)


exjam:
So I had a play with this a couple times over the past few days, and unfortunately it seems h264.rpl is a bit more complex than this - I was testing with MK8 which seems to reply on h264.rpl for parsing the SPS and ensuring they have full NAL units to render.

**H264DECCheckDecunitLength:**
Check that we have a full NALU in buffer which we can decode.

**H264DECGetImageSize:**
Parse SPS NALU and extract image width and height.

**H264DECFindDecstartpoint**
Find next SPS NALU.

**H264DECFindIdrpoint:**
Find next IDR frame NALU.

It would seem they basically read from a raw h264 stream from file and use h264.rpl to ensure they have read enough data to decode:
```
while (!H264DECFindDecstartpoint) {
  FSReadFile through the file trying to find start point
}
H264DECBegin
H264DECGetImageSize

while (playing) {
  while (!H264DECCheckDecunitLength) {
    FSReadFile until we have a full NALU
  }
  H264DECSetBitstream
  H264DECExecute
}
```

H264DECFindDecstartpoint is trivial because finding a NALU SPS is really easy, but the other methods require fully parsing the h264 stream.

I really struggled to get ffmpeg to parse the h264 stream and give the data I need, for example in H264DECGetImageSize we only have the SPS, so i tried passing the SPS in the AVContext extradata, and it does parse it but ffmpeg will not update the context ->width and ->height until a frame is flushed from the decoder. I know for sure it actually is parsing it though because if you set the debug log callback it will call it with a string containing info from the sps including the macroblock width & height.

I ended up making my own SPS bitstream parser, which took a while but it worked to get the width and height for GetImageSize. But then I looked at H264DECCheckDecunitLength and got depressed - as it requires fully parsing the stream in order to check if we have a full NALU which is much more work than just a SPS parser.

So basically either we figure out how to get ffmpeg to do this work for us, use a different library or write our own h264 bitstream decoder. Was wondering what your thoughts are on this?

You can see what I did here btw:
https://github.com/exjam/decaf-emu/blob/h264/src/libdecaf/src/cafe/libraries/h264/h264_dec.cpp
https://github.com/exjam/decaf-emu/commit/7c2ed0ea2ec98e098ecf4aa69778b5444fa1a3f0

VelocityRa:
I take it it's not possible to LLE `h264.rpl`?

exjam:
It is possible - but what about people who do not have a h264.rpl ? :)

SteveLeafo:
Hey - I never thought to test with Mario Kart, I was just trying to get The Smurfs to run :) 

I am happy to keep fiddling with ffmpeg to see if I can get Mario Kart working.   Writing our own h264 decoder sounds fun, but I guess it would be difficult to match the performance of the existing libraries out there.

I am just finishing off some changes to padscore (to support multiple controllers) so I might come back to this in a little while.

I am also keen to have a look at the audio issues - do you have any ideas on why it's so crackly?

SteveLeafo:
Unrelated to this pull, I have added support for Decaf to Budford. 

It makes testing changes on multiple games a lot easier and has the ability to stop decaf when it gets stuck.

image

VelocityRa:
@exjam Sent you a response on IRC to not sidetrack from this convo.

SteveLeafo:
I had a scrounge around and came across this guy:
https://sourceforge.net/p/h264bitstream/code/HEAD/tree/trunk/h264bitstream/

It's LGPL - not sure if that's a problem.

It seems like it's a complete h264 bitstream parser - though it may be a little old.

I plumbed it in and it seems to be getting all the right information and I have had Mario Kart successfully decode an entire stream (Well - it decoded lots of packets without errors and called H264DECClose at the end).

I'll update this pull request for all to see.

Was a bit of a let down as the video didn't end up displaying after all.

exjam:
Yeah I found https://github.com/aizvorski/h264bitstream/ which looks like the more updated version of the one you linked.

SteveLeafo:
So, when I run this code it always ends up with something like:
[05:58:23.093400 error:p1 t25] COSVReport1 0: [OMXCodec] [OMX.nerd.h264.decoder] Timed out waiting for output buffers: 0/1
I am pretty sure the data being sent to the call back function is sensible, but I was never able to get it to send a second chunk.

exjam:
Thanks for your help and research on this, I combined our work and ended up with daa17066f3f8fa78b5e9fa66809b64ce7c557460

ffmpeg is optional, if decaf is built without ffmpeg then it will output fake frames like:
https://puu.sh/C8vM9/153fc82adf.jpg

Seems to work for Smurfs at least!

PR #496, opened 2018-10-05
Implementation of GX2SubmitUserTimeStamp

SteveLeafo:
This change makes quite a few games playable and allows others to progress a little further

### DIFF
--- a/src/libdecaf/src/cafe/libraries/gx2/gx2_enum.h
+++ b/src/libdecaf/src/cafe/libraries/gx2/gx2_enum.h
@@ -338,6 +338,12 @@ ENUM_BEG(GX2StreamOutContextMode, uint32_t)
    ENUM_VALUE(FromOffset,              2)
 ENUM_END(GX2StreamOutContextMode)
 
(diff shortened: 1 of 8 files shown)


exjam:
GX2SubmitUserTimestamp actually submits PM4 commands which the GPU executes to write back the user timestamp.

For GX2PipeEvent::Top it will use a MEM_WRITE, for Bottom it uses EVENT_WRITE_OP with BOTTOM_OF_PIPE_TS. and BottomAfterFlush it uses an EVENT_WRITE_EOP with CACHE_FLUSH_AND_INV_TS_EVENT.

See gx2_query.cpp for an example of sending a MEM_WRITE and an EVENT_WRITE_EOP:
https://github.com/decaf-emu/decaf-emu/blob/master/src/libdecaf/src/cafe/libraries/gx2/gx2_query.cpp#L19-L52

exjam:
Thanks, merged as 13eef6fb7462143b9eaddf6b2890de4d55ae2fa1

PR #497, opened 2018-10-07
Implementation of KPAD / WPAD

SteveLeafo:
With these changes I was able to play a two player game of family tennis, but there is still a long way to go.
I wasn't able to get a second player to join a Mario Kart game.


src/decaf-sdl/decafsdl_input.cpp
```diff
@@ -523,5 +758,47 @@ DecafSDL::getAxisValue(wpad::Channel channel, wpad::NunchuckAxis axis)
...
+   }
+
+   case config::input::Joystick:
+      if (mWpadController[(int)channel] && SDL_GameControllerGetAttached(mWpadController[(int)channel])) {
+         auto value = getJoystickAxisState(mWpadConfig[(int)channel], mWpadController[(int)channel], channel, axis);
+
+         if (value < 0) {
+            return value / 32768.0f;
+         }
+         else {
```
VelocityRa: `} else {`

src/libdecaf/src/cafe/libraries/padscore/padscore_wpad.cpp
```diff
@@ -183,6 +199,10 @@ WPADControlDpd(int32_t chan,
    return WPADError::OK;
 }
 
+bool IsProControllerAllowed()
```
VelocityRa: return type on separate line

src/libdecaf/src/cafe/libraries/padscore/padscore_kpad.h
```diff
@@ -89,6 +100,26 @@ struct KPADExtStatus
 };
 CHECK_SIZE(KPADExtStatus, 0x50);
 
+ struct KPADMPDir {
```
VelocityRa: remove space before `struct`

exjam:
Looking pretty solid overall - will merge it when I get a chance at my PC.

I was wondering if you had any thoughts on improving the API between the front-end and decaf itself? I'm not so sure the API I threw together the first time round was a great choice.

Maybe instead of having a function per button in the decaf InputDriver we should consider consolidating them into basically how VPADRead works and read the structure in one go?

Thinking ahead too I think we are also going to need support for frontend notifying decaf of controllers being connected and other such controller events, also adding methods for vibration etc.

Although I am not sure where this could fit in if we ever did bluetooth / usb passthrough as the games do have access to USB devices.

Something to consider for a future PR at least - if you have any ideas about a nicer API that is.

exjam:
padscore_kpad.cpp gCoreButtonMap is now unused.

In KPADReadEx you use gLastButtonState but don't update its value., Also gLastButtonState should be moved to sKpadData->lastButtonState. And it needs to be per controller rather than for all.

So something roughly like this:
```
auto lastButtonState = sKpadData->lastButtonState[channel];
auto currentButtonState = uint32_t { 0 };

....

auto previous = lastButtonState & bit;
if (status == input::ButtonStatus::ButtonPressed) {
currentButtonState |= bit;
}

...

sKpadData->lastButtonState[channel] = currentButtonState;
```

src/common-sdl/decafsdl_config.cpp
```diff
@@ -266,6 +268,16 @@ loadFrontendToml(std::shared_ptr<cpptoml::table> config)
...
+
+   config::input::wpad[0] = config->get_qualified_as<std::string>("input.wpad0").value_or(config::input::wpad[0]);
+   config::input::wpad[1] = config->get_qualified_as<std::string>("input.wpad1").value_or(config::input::wpad[1]);
+   config::input::wpad[2] = config->get_qualified_as<std::string>("input.wpad2").value_or(config::input::wpad[2]);
+   config::input::wpad[3] = config->get_qualified_as<std::string>("input.wpad3").value_or(config::input::wpad[3]);
+
+   config::input::wpadType[0] = config->get_qualified_as<std::string>("input.wpadType0").value_or(config::input::wpadType[0]);
+   config::input::wpadType[1] = config->get_qualified_as<std::string>("input.wpadType1").value_or(config::input::wpadType[1]);
+   config::input::wpadType[2] = config->get_qualified_as<std::string>("input.wpadType2").value_or(config::input::wpadType[2]);
+   config::input::wpadType[3] = config->get_qualified_as<std::string>("input.wpadType3").value_or(config::input::wpadType[3]);
```
baryluk: change these repetitive code to use for loops?

also in other places below and in config.cpp and main.cpp ?

brett19: It would require using string formatting to do that, probably nicer like it is here?

baryluk:
```cpp
   for (int i = 0; i < 4; i++) {
       config::input::wpadType[i] = config->get_qualified_as<std::string>("input.wpadType" + std::to_string(i)).value_or(config::input::wpadType[i]);
   }
```

or simply `std::to_string`, no need to really use anything more fancy like `fmt`. Nothing wrong with some small formatting or use of a lookup table (`std:string[4] digits = {"0", "1", "2", "3"};`).

brett19: Fair point (👍)

PR #498, opened 2018-10-21
wiiu-rpc: Fix compatibility for newer wut versions

QuarkTheAwesome:
Heyo!
So it turns out that if you have a machine set up with wut, `-DDECAF_BUILD_TOOLS=ON` will always attempt to build wiiu-rpc. This commit fixes up that code to build with newer revisions of wut. I tested it, and the rpx actually works too! May come in handy, I guess.
Thanks!
Ash


exjam:
Merged as eb23911c1e4ec54c3054c68b7ef8c4cb6aa6cfd7

PR #503, opened 2018-11-30
vulkan: Minor compilation fixes and one possible bug

baryluk:
vulkan: Minor compilation fixes and one possible bug

Do not take address of temporaries.
Use proper casts when casting from Vulkan-Hpp to Vulkan.
Fix use of implicit conversions that make gcc not like it
due to overload ambiguity.
Remove unused bool cast operator, that called non existing
bool operator in other class.

Tested with vulkan 1.1.82 and 1.1.73, and gcc 8.2.

### DIFF
--- a/src/libgpu/src/spirv/spirv_cf.cpp
+++ b/src/libgpu/src/spirv/spirv_cf.cpp
@@ -170,18 +170,22 @@ void Transpiler::translateCf_LOOP_START_DX10(const ControlFlowInst &cf)
    auto stateVal = mSpv->createLoad(mSpv->stateVar());
 
    // If state is set to InactiveContinue, set it to Active
-   auto isInactiveContinue = mSpv->createBinOp(spv::OpIEqual, mSpv->boolType(),
-                                               stateVal,
-                                               mSpv->stateInactiveContinue());
-   auto ifBuilder = &spv::Builder::If { isInactiveContinue, spv::SelectionControlMaskNone, *mSpv };
-   mSpv->createStore(mSpv->stateActive(), mSpv->stateVar());
-   ifBuilder->makeEndIf();
+   {
+      auto isInactiveContinue = mSpv->createBinOp(spv::OpIEqual, mSpv->boolType(),
+                                                  stateVal,
+                                                  mSpv->stateInactiveContinue());
+      auto ifBuilder = spv::Builder::If { isInactiveContinue, spv::SelectionControlMaskNone, *mSpv };
+      mSpv->createStore(mSpv->stateActive(), mSpv->stateVar());
+      ifBuilder.makeEndIf();
+   }
 
    // If state is set to Inactive, set it to InactiveBreak
-   auto isInactive = mSpv->createBinOp(spv::OpIEqual, mSpv->boolType(), stateVal, mSpv->stateInactive());
-   ifBuilder = &spv::Builder::If { isInactive, spv::SelectionControlMaskNone, *mSpv };
-   mSpv->createStore(mSpv->stateInactiveBreak(), mSpv->stateVar());
-   ifBuilder->makeEndIf();
+   {
+      auto isInactive = mSpv->createBinOp(spv::OpIEqual, mSpv->boolType(), stateVal, mSpv->stateInactive());
+      auto ifBuilder = spv::Builder::If { isInactive, spv::SelectionControlMaskNone, *mSpv };
+      mSpv->createStore(mSpv->stateInactiveBreak(), mSpv->stateVar());
+      ifBuilder.makeEndIf();
+   }
 }
 
 void Transpiler::translateCf_LOOP_START_NO_AL(const ControlFlowInst &cf)
--- a/src/libgpu/src/vulkan/vulkan_driver.h
+++ b/src/libgpu/src/vulkan/vulkan_driver.h
@@ -13,6 +13,7 @@
 #include <condition_variable>
 #include <functional>
 #include <gsl/gsl>
+#include <list>
 #include <map>
 #include <mutex>
 #include <thread>
@@ -48,11 +49,6 @@ public:
    DataHash hash() const
    {
       return mHash;
-   }
-
-   operator bool() const
-   {
-      return !!mHash;
    }
 
    bool operator==(const HashedDesc<T>& other) const
--- a/src/libgpu/src/vulkan/vulkan_staging.cpp
+++ b/src/libgpu/src/vulkan/vulkan_staging.cpp
@@ -31,7 +31,7 @@ Driver::allocTempBuffer(uint32_t size)
    VkBuffer buffer;
    VmaAllocation allocation;
    vmaCreateBuffer(mAllocator,
-                   &static_cast<VkBufferCreateInfo>(bufferDesc),
+                   reinterpret_cast<VkBufferCreateInfo*>(&bufferDesc),
                    &allocDesc,
                    &buffer,
                    &allocation,
--- a/src/libgpu/src/vulkan/vulkan_surface.cpp
+++ b/src/libgpu/src/vulkan/vulkan_surface.cpp
@@ -181,24 +181,24 @@ Driver::_copySurface(SurfaceObject *dst, SurfaceObject *src, SurfaceSubRange ran
 
    vk::ImageCopy copyRegion;
    if (src->desc.dim == latte::SQ_TEX_DIM::DIM_3D) {
-      copyRegion.srcOffset = { 0, 0, static_cast<int32_t>(range.firstSlice) };
+      copyRegion.srcOffset = vk::Offset3D { 0, 0, static_cast<int32_t>(range.firstSlice) };
       copyRegion.srcSubresource = { copyAspect, 0, 0, 1 };
       // range.numSlices is expressed by the extent here...
    } else {
-      copyRegion.srcOffset = { 0, 0, 0 };
+      copyRegion.srcOffset = vk::Offset3D { 0, 0, 0 };
       copyRegion.srcSubresource = { copyAspect, 0, range.firstSlice, range.numSlices };
    }
    if (dst->desc.dim == latte::SQ_TEX_DIM::DIM_3D) {
-      copyRegion.dstOffset = { 0, 0, static_cast<int32_t>(range.firstSlice) };
+      copyRegion.dstOffset = vk::Offset3D { 0, 0, static_cast<int32_t>(range.firstSlice) };
       copyRegion.dstSubresource = { copyAspect, 0, 0, 1 };
       // range.numSlices is expressed by the extent here...
    } else {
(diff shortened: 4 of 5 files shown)


src/libgpu/src/vulkan/vulkan_staging.cpp
```diff
@@ -28,10 +30,13 @@ Driver::allocTempBuffer(uint32_t size)
    VmaAllocationCreateInfo allocDesc = {};
    allocDesc.usage = VMA_MEMORY_USAGE_CPU_TO_GPU;
 
+
    VkBuffer buffer;
    VmaAllocation allocation;
+   static_assert(sizeof(vk::BufferCreateInfo) == sizeof(VkBufferCreateInfo),
```
brett19: This is guaranteed by the Vulkan-Hpp spec already.

src/libgpu/src/vulkan/vulkan_surface.cpp
```diff
@@ -181,24 +181,24 @@ Driver::_copySurface(SurfaceObject *dst, SurfaceObject *src, SurfaceSubRange ran
 
    vk::ImageCopy copyRegion;
    if (src->desc.dim == latte::SQ_TEX_DIM::DIM_3D) {
-      copyRegion.srcOffset = { 0, 0, static_cast<int32_t>(range.firstSlice) };
+      copyRegion.srcOffset = VkOffset3D { 0, 0, static_cast<int32_t>(range.firstSlice) };
```
brett19: Interesting.  This shouldn't be needed I don't think?  Why does it fail?

src/libgpu/src/vulkan/vulkan_driver.h
```diff
@@ -50,11 +51,6 @@ class HashedDesc
       return mHash;
    }
 
-   operator bool() const
```
brett19: Why is this removed here?

src/libgpu/src/spirv/spirv_cf.cpp
```diff
@@ -170,18 +170,22 @@ void Transpiler::translateCf_LOOP_START_DX10(const ControlFlowInst &cf)
...
+   }
 
    // If state is set to Inactive, set it to InactiveBreak
-   auto isInactive = mSpv->createBinOp(spv::OpIEqual, mSpv->boolType(), stateVal, mSpv->stateInactive());
-   ifBuilder = &spv::Builder::If { isInactive, spv::SelectionControlMaskNone, *mSpv };
-   mSpv->createStore(mSpv->stateInactiveBreak(), mSpv->stateVar());
-   ifBuilder->makeEndIf();
+   {
+      auto isInactive = mSpv->createBinOp(spv::OpIEqual, mSpv->boolType(), stateVal, mSpv->stateInactive());
+      auto ifBuilder = spv::Builder::If { isInactive, spv::SelectionControlMaskNone, *mSpv };
```
brett19: Oops shit.  Nice catch.

src/libgpu/src/vulkan/vulkan_driver.h
baryluk: It is unused method. What would be even the meaning of bool(hash) ?

src/libgpu/src/vulkan/vulkan_surface.cpp
baryluk:
```
/home/user/decaf-emu/src/libgpu/src/vulkan/vulkan_surface.cpp: In member function ‘vulkan::SurfaceObject* vulkan::Driver::_allocateSurface(const vulkan::SurfaceDesc&)’:
/home/user/decaf-emu/src/libgpu/src/vulkan/vulkan_surface.cpp:480:41: error: ambiguous overload for ‘operator=’ (operand types are ‘vk::Offset3D’ and ‘<brace-enclosed initializer list>’)
    bufferRegion.imageOffset = { 0, 0, 0 };
                                         ^
In file included from /home/user/decaf-emu/src/libgpu/./gpu_vulkandriver.h:4,
                 from /home/user/decaf-emu/src/libgpu/src/vulkan/vulkan_driver.h:5,
                 from /home/user/decaf-emu/src/libgpu/src/vulkan/vulkan_surface.cpp:2:
/usr/include/vulkan/vulkan.hpp:4948:15: note: candidate: ‘vk::Offset3D& vk::Offset3D::operator=(const VkOffset3D&)’
     Offset3D& operator=( VkOffset3D const & rhs )
               ^~~~~~~~
/usr/include/vulkan/vulkan.hpp:4925:10: note: candidate: ‘constexpr vk::Offset3D& vk::Offset3D::operator=(const vk::Offset3D&)’
   struct Offset3D
          ^~~~~~~~
/usr/include/vulkan/vulkan.hpp:4925:10: note: candidate: ‘constexpr vk::Offset3D& vk::Offset3D::operator=(vk::Offset3D&&)’

```

Am I using to old libvulkan?

src/libgpu/src/spirv/spirv_cf.cpp
baryluk: Yeah, taking an address of the temporary is a bad idea.

src/libgpu/src/vulkan/vulkan_staging.cpp
baryluk: Indeed, you are right. I even see in Vulkan-Hpp


  static_assert( sizeof( BufferCreateInfo ) == sizeof( VkBufferCreateInfo ), "struct and wrapper have different size!" );

So, this is redundant, and all structs are already checked.

src/libgpu/src/vulkan/vulkan_driver.h
brett19: It used to be used to verify if there is actually a hash stored (the hash uint64_t[2]{0, 0}).  I think I ended up getting rid of it in exchange for checking if the last usage is non-zero.

src/libgpu/src/vulkan/vulkan_staging.cpp
baryluk: Removed.

src/libgpu/src/vulkan/vulkan_surface.cpp
brett19: I'm using the latest SDK (effectively, I custom built it for transform feedback).  That's a stupid warning though, since the ambiguous overload it mentions are for the same thing....  They both initialize the same vk type >_<.

brett19:
Could you squash the commits together (as we only do rebase merges) and append the `vulkan:` prefix to the commit message to fit with the commit message styling already in use by the project?

src/libgpu/src/vulkan/vulkan_surface.cpp
brett19: In hindsight, I don't think the extra specificity is actually a bad thing.  You will note that in almost every case, I try to avoid using initializer lists anyways, since the parameters being set are so unclear when you read.

baryluk: I checked my copy in of Offset3D constructors and the one in master branch of Vulkan-Hpp, and they look identical:

    Offset3D( int32_t x_ = 0,
              int32_t y_ = 0,
              int32_t z_ = 0 );
    explicit Offset3D( Offset2D const& offset2D,
                       int32_t z_ = 0 );
    Offset3D( VkOffset3D const & rhs );

and operator = is also the same:

    Offset3D& operator=( VkOffset3D const & rhs );


so I think it is not related to version of vulkan.

The problem is that the compiler do not know which version to call operator=( VkOffset3D const & rhs ); or compiler provided operator= methods.

This is because due to implicit argument type conversion. the brace-enclosed init list can both be passed to constructors of VkOffset3D and vk::Offset3D.  (this is because VkOffset3D has trivial constructor that can do it).

Yes, the bodies and semantic are the exactly same, but compiler doesn't know that. So you just need to select one you prefer.

I can make it vk::Offset3D { ... }.  I guess it might be more consistent with rest of the code.

I just tested it and it also work.

brett19: Yup.  Use vk::Offset3D and it will be good.  Then just the commit squash / message and I can merge once CI finishes :)

baryluk:
squashed and edited the commit message.

I do not like doing `git push -f`, but `git rebase -i` + squash looks awesome. Never tried it before I think.

brett19:
Yea, usually you would avoid doing a force push, but in this case, since its a PR specifically, it makes a lot of sense to clean it up before we push it to the 'canonical' repository :)

src/libgpu/src/vulkan/vulkan_surface.cpp
brett19: here

src/libgpu/src/vulkan/vulkan_surface.cpp
```diff
@@ -181,24 +181,24 @@ Driver::_copySurface(SurfaceObject *dst, SurfaceObject *src, SurfaceSubRange ran
...
       copyRegion.srcSubresource = { copyAspect, 0, 0, 1 };
       // range.numSlices is expressed by the extent here...
    } else {
-      copyRegion.srcOffset = { 0, 0, 0 };
+      copyRegion.srcOffset = VkOffset3D { 0, 0, 0 };
```
brett19: here

src/libgpu/src/vulkan/vulkan_surface.cpp
```diff
@@ -181,24 +181,24 @@ Driver::_copySurface(SurfaceObject *dst, SurfaceObject *src, SurfaceSubRange ran
...
       copyRegion.srcSubresource = { copyAspect, 0, range.firstSlice, range.numSlices };
    }
    if (dst->desc.dim == latte::SQ_TEX_DIM::DIM_3D) {
-      copyRegion.dstOffset = { 0, 0, static_cast<int32_t>(range.firstSlice) };
+      copyRegion.dstOffset = VkOffset3D { 0, 0, static_cast<int32_t>(range.firstSlice) };
```
brett19: here

src/libgpu/src/vulkan/vulkan_surface.cpp
```diff
@@ -181,24 +181,24 @@ Driver::_copySurface(SurfaceObject *dst, SurfaceObject *src, SurfaceSubRange ran
...
       copyRegion.dstSubresource = { copyAspect, 0, 0, 1 };
       // range.numSlices is expressed by the extent here...
    } else {
-      copyRegion.dstOffset = { 0, 0, 0 };
+      copyRegion.dstOffset = VkOffset3D { 0, 0, 0 };
```
brett19: and here

src/libgpu/src/vulkan/vulkan_surface.cpp
baryluk: Done

brett19:
Didn't bother waiting for CI, since CI doesn't check the Vulkan stuff yet anyways.

PR #505, opened 2018-12-02
Changes to allow decaf-cli to cleanly exit

CreeperMario:
While trying to run the test suite changes in my other PR, I found that `decaf-cli` was exiting before the test program had even started running, and was not spitting out any error messages.

I found that this was because `decaf::waitForExit()` calls `ios::join()` and `cafe::kernel::join()` which signal the respective module threads to start shutting down, before they have even started running the program.

This would also result in a crash because `decaf::waitForExit()` used to call `decaf::shutdown()`, which also tries to join the IOS/Kernel threads, resulting in `std::__1::system_error` being thrown because an already terminated thread cannot be joined.

I implemented new exit functionality to halt the main thread until `sStopping` in `src/libdecaf/src/cafe/kernel/cafe_kernel.cpp` is set to `true` using an `std::condition_variable`.

Because a condition variable is now used to exit `decaf-cli` upon the emulated program's return, the process of exiting `decaf-cli` upon a specified timeout can be much more cleanly implemented as well. This new timeout code also solves an `std::__1::system_error` I was experiencing due to a fault in `std::mutex` usage that I couldn't locate.

This is my first time working with C++ threading features, so feedback is appreciated. 🙂

### DIFF
--- a/src/decaf-cli/decafcli.cpp
+++ b/src/decaf-cli/decafcli.cpp
@@ -35,13 +35,13 @@ DecafCLI::run(const std::string &gamePath)
    std::condition_variable timeoutCV;
 
    if (config::system::timeout_ms) {
-      std::mutex timeoutMutex;
       auto timeout = std::chrono::system_clock::now() + std::chrono::milliseconds(config::system::timeout_ms);
 
       timeoutThread = std::thread {
          [&]() {
+            std::mutex timeoutMutex;
             std::unique_lock<std::mutex> lock { timeoutMutex };
-            timeoutCV.wait_until(lock, timeout, [&]() { return running.load(); });
+            timeoutCV.wait_until(lock, timeout, [&]() { return !running.load(); });
 
             if (running) {
                timedOut.store(true);
--- a/src/libdecaf/src/cafe/kernel/cafe_kernel.cpp
+++ b/src/libdecaf/src/cafe/kernel/cafe_kernel.cpp
@@ -286,8 +286,6 @@ stop()
       sStopping = true;
       cpu::halt();
    }
-
-   cpu::join();
 }
 
 void
--- a/src/libdecaf/src/cafe/kernel/cafe_kernel.cpp
+++ b/src/libdecaf/src/cafe/kernel/cafe_kernel.cpp
@@ -318,6 +318,14 @@ idleCoreLoop(cpu::Core *core)
 void
 exit()
 {
+   // Cafe kernel is about to exit - IOS threads should also stop.
+   auto error = IOS_Ioctl(RamPartitionId::Kernel,
+                          RamPartitionId::Invalid,
+                          getPpcAppHandle(),
+                          ios::mcp::PPCAppCommand::PowerOff,
+                          nullptr, 0,
+                          nullptr, 0);
+
    // Set the running flag to false so idle loops exit.
    sStopping = true;
 
--- a/src/libdecaf/src/cafe/kernel/cafe_kernel.cpp
+++ b/src/libdecaf/src/cafe/kernel/cafe_kernel.cpp
@@ -276,6 +276,12 @@ hasExited()
 void
 join()
 {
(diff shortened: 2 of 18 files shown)


exjam:
Sorry for taking an unreasonable amount of time to get around to this - I was selfishly putting off looking at the more complicated PRs for a while whilst I finished my own things.

I think the real problem here is that I screwed up ios::join and cafe::kernel::join when I did my big rewrite - the definition of join is that it should sit and wait until the threads finish exeuction. Instead I incorrectly implemented join more like a shutdown/stop method.

wrt to "This is my first time working with C++ threading features, so feedback is appreciated. " - you did a perfect job with them to achieve your goal and I have no feedback to offer you!

However I think we need to look into fixing all the join methods to actually do a join and add separate stop methods for when we actually want a stop. I imagine it would look something like this:

ios::join would need to first call a new method kernel::join which does a proper join, and then we can renmae the existing internal::joinXThread to stopWorkerThread and stopAlarmThread and call those (because we do actually want to stop these once the kernel thread has exited - even in this ios::join method).

Also we would add an ios::stop which called kernel::stop followed by the same stopWorkerThread and stopAlarmThread as we did in ios::join.

cafe::kernel::join would just call cpu::join without setting the sStopping flag and then we'd have a cafe::kernel::stop which does same thing as the existing join.

Let me know if you're still interested in taking a stab at doing that, or I can do it myself.

CreeperMario:
I just pushed to a separate branch [`cli-exit-attempt-3`](https://github.com/CreeperMario/decaf-emu/tree/cli-exit-attempt-3) which implements your ideas (moving shutdown code to dedicated `::stop()` functions).

It seems fine in theory, except with these changes, `decaf-cli` no longer exits at all, and I have a theory as to why that is.

When the RPX returns, the PPC threads exit automatically, whereas the IOS hardware thread keeps running until it is instructed to stop from an external source (i.e. it will not stop when the RPX returns). As such, trying to join to the IOS kernel from the main thread (as `decaf-cli` does) will eventually result in a hang as the main thread is now waiting for the hardware thread to exit, which never happens.

A possible method of fixing this would be to have the PPC kernel thread appropriately set the hardware thread's `sRunning` to `false` to make it begin exiting, but I am unsure of the best method of implementing this.

Also, the branch was created from the commit before VFS was implemented, because guess which C++17 feature Xcode doesn't support? If you'd rather I base the changes off of a post-VFS commit, I'm happy to do so but I won't be able to test the changes until I finish setting up my Arch distro.

exjam:
OK yeah what we need to do next then is have IOS notified when the PPC shutdowns - the kernel actually sends an ioctl with command 0xB2 to /dev/ppc_app on shutdown (which is read here: https://github.com/decaf-emu/decaf-emu/blob/master/src/libdecaf/src/ios/mcp/ios_mcp_ppc_thread.cpp)

So we can read this ioctl and consequently shutdown IOS which should hopefully lead to decaf-cli exiting!

You can see hex-rays output for the ppc thread function here:
https://gist.github.com/exjam/4f1db094aabc8ee723353b2e2736ab11#file-mcp-ppc-thread-L204-L229

And yes sorry Apple is quite behind on the C++ front. It might be possible to download a more modern clang and libstdc++ using brew or something I do not know.

CreeperMario:
`decaf-cli` now exits normally!

All that's left now is fixing exiting on a timeout (using config option `timeout_ms`) which still causes the following crash:

```
libc++abi.dylib: terminating with uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument
Abort trap: 6
```

I'll see if I can find a solution tomorrow.

CreeperMario:
Done, but my branch now has a conflict with master. Do you want me to solve it or are you happy to merge locally and then push?

exjam:
Thanks for fixing this.

CreeperMario:
No worries, thanks for sharing your advice!

PR #506, opened 2018-12-02
Updates to test suites and RPC server/client

CreeperMario:
The grunt of this PR is that the RPC client and HLE test cases were updated so that they can be built with WUT versions past `203dc66` (the current `legacy_cmake` branch)

In the case of the RPC client, I also updated the example `client.py` script such that the output of `MCP_TitleList` is printed out on screen more cleanly.

I also fixed some compiler errors in the CPU test cases which were likely the result of recent changes in `libcpu`.

And I changed the `alarm_multithread` test case such that, instead of using the default threads on cores 0 and 2 to run code, new threads are created. I did this because `OSJoinThread`, when used with the default core threads, seems to immediately return `FALSE` on-console, and throws an assertion failure error in Decaf. Using our own threads instead seems to resolve the error.

When combined with my other PR, all HLE test cases run and pass on my system, and almost all CPU tests pass as well.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -42,7 +42,7 @@ if(DECAF_BUILD_WUT_TESTS)
         CMAKE_CACHE_ARGS
             -DDEVKITPPC:string=${DEVKITPPC}
             -DWUT_ROOT:string=${WUT_ROOT}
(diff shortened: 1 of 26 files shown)


CreeperMario:
No, I didn't see #498. Oops. 😇

exjam:
Looks good, thanks :).

PR #507, opened 2018-12-02
cafe/hle: Properly initialize LibraryFunction fields

baryluk:
Most notably hostPtr might contain garbage, that will subsequently
lead to write to random memory location when doing relocations.

As a minor additional tweak, initialize LibraryFunctionCall::func
in constructor.

It took me a lot of time to find this out. I was getting consistent
crashes when trying to reloc ACConnect function, but couldn't
figure out easily how the random values got into hostPtr field.
Finally I tracked it with help of gdb and some dynamic watchpoints.

### DIFF
--- a/src/libdecaf/src/cafe/libraries/cafe_hle_library_function.h
+++ b/src/libdecaf/src/cafe/libraries/cafe_hle_library_function.h
@@ -13,9 +13,9 @@ namespace cafe::hle
 
 struct UnimplementedLibraryFunction
 {
-   class Library *library;
+   class Library *library = nullptr;
    std::string name;
-   uint32_t syscallID;
+   uint32_t syscallID = 0;
    virt_addr value;
 };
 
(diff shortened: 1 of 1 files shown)


exjam:
I think syscallID == 0 might be a valid syscall value, so we should probably avoid initialising it to 0 and do -1 instead or something?

baryluk:
I guess so. I was researching the cast from -1 to unsigned, because it feels to be undefined, and according to spec (C and C++ right now) it is undefined (`UINT32_MAX` is probably better and more portable), but apparently all machines in the world use two complement signed integers these days, and all compilers do support it, and there is even a C++ draft proposal to make signed to unsigned casts, and signed overflows, fully defined. (As opposed to current C standard). So, -1 isn't that bad.

So, yeah, `-1` should do a trick.

Will push code later.

exjam:
I changed it to -1: 0edbcf9667fd70ee488d3b08fe261d2f375d7fc5

PR #509, opened 2018-12-05
vulkan: Don't take the address of temporaries.

achurch:
GCC treats this as an error.


brett19:
`VkBufferCreateInfo` and `vk::BufferCreateInfo` are guaranteed to be the same, `reinterpret_cast<VkBufferCreateInfo*>(&vkBufferDesc)` should be used instead (that's what we do elsewhere).

achurch:
Changed.

brett19:
Thanks @achurch!

PR #515, opened 2019-01-03
Changes to make DKTF go in game

SteveLeafo:
Small change to OpenGL implementation to allow games with Geometry shaders to progress.


exjam:
This was intentional, we do not want to ignore problems.

PR #521, opened 2019-01-06
Changes to ensure settings are saved in decaf-qt

SteveLeafo:


exjam:
Might aswell just call saveSettings immediately, avoids the settings being lost if decaf crashes.

https://github.com/decaf-emu/decaf-emu/commit/7abeac40977e9be0674c2b350cbf362d3919fbba

PR #523, opened 2019-01-06
 Fixed a crash that occurred when applying settings.

SteveLeafo:
When I clicked the apply settings buttons, decaf would crash if it failed to parse my config file.  When this happens I just create a new table.


exjam:
Ah yeah I should have tested it properly...

PR #524, opened 2019-01-06
Shader binaries dumping support

VelocityRa:
* Intentionally only touched the Vulkan backend, only works there.
* New config option to _only_ dump binaries (instead of disassemblies and such). Reasoning is that we need to see binaries of shaders that decaf's latte disassembler doesn't understand yet.


exjam:
483d2bbb8a8f84643372ea849cf3fc12831cd922
2666a1153463fa77174b1dbbf328f62e5bdaa2b3

PR #527, opened 2019-01-11
Installation changes

Tatsh:
This makes the installation and path handling a little more normal for a system install.

* All binaries including tools go in `${prefix}/bin` (normally `/usr/bin`)
* `-DDECAF_RESOURCES_PATH=/some-path/` can be used to set the default resources path
* If `DECAF_RESOURCES_PATH` is not set, all resources go to `${prefix}/share/decaf-emu/resources` (normally `/usr/share/...`)
* All documentation goes in `${prefix}/share/docs/decaf-emu`
* Although maybe useless, it is possible to build with `-DDECAF_QT=ON` and `DDECAF_SDL=OFF`. Just for consistency

### DIFF
--- a/resources/CMakeLists.txt
+++ b/resources/CMakeLists.txt
@@ -1,3 +1,5 @@
+include(GNUInstallDirs)
+
 project(resources)
 
 set(Fonts
@@ -5,9 +7,13 @@ set(Fonts
    fonts/CafeKr.ttf
    fonts/CafeStd.ttf
    fonts/CafeTw.ttf
-   fonts/NotoSansCJK.LICENSE
-   fonts/DejaVuSansMono.ttf
-   fonts/DejaVuSansMono.LICENSE)
+   fonts/DejaVuSansMono.ttf)
+set(Docs
+    fonts/DejaVuSansMono.LICENSE
+    fonts/NotoSansCJK.LICENSE
+    ../README.md
+    ../LICENSE.md)
+set(PARENT_PROJECT_NAME decaf-emu)
 
 set(ResourceFiles ${Fonts})
 
@@ -20,5 +26,13 @@ foreach(ResourceFile ${ResourceFiles})
                           $<TARGET_FILE_DIR:libdecaf>/resources/${ResourceFile})
 endforeach()
 
-install(FILES ${Fonts}
-        DESTINATION "${CMAKE_INSTALL_PREFIX}/resources/fonts")
+if(DECAF_RESOURCES_PATH)
+  install(FILES ${Fonts} DESTINATION "${DECAF_RESOURCES_PATH}")
+else()
+  install(FILES ${Fonts}
+          DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/${PARENT_PROJECT_NAME}/resources/fonts")
+endif()
+
+if(DECAF_DOCS)
+  install(FILES ${Docs} DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/doc/${PARENT_PROJECT_NAME}")
+endif()
--- a/src/decaf-cli/CMakeLists.txt
+++ b/src/decaf-cli/CMakeLists.txt
@@ -1,3 +1,5 @@
+include(GNUInstallDirs)
+
 project(decaf-cli)
 
 include_directories(".")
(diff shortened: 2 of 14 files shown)


CreeperMario:
> Although maybe useless, it is possible to build with -DDECAF_QT=ON and DDECAF_SDL=OFF. Just for consistency

Isn't SDL still used for input? Or did I miss something?

Tatsh:
Yes, that's why I wrote that a build with that combination is probably useless. In that scenario, `libcommon-sdl.a` is not generated, so it's not possible to link `decaf-qt`.

If the options are there, it seems like any combination should work, or the combination should fail and alert the user to the problem (e.g. *You cannot build the Qt binary without SDL enabled*).

exjam:
In it's current state won't this mean that unless you specify DECAF_RESOURCES_PATH then it will default the config to have resources path simply as "resources" and hence will not be looking in ${prefix}/share/decaf-emu/resources ?

Tatsh:
> In it's current state won't this mean that unless you specify DECAF_RESOURCES_PATH then it will default the config to have resources path simply as "resources" and hence will not be looking in ${prefix}/share/decaf-emu/resources ?

Yes. I wasn't sure if the default behaviour should be changed here. If it's okay, I can remove the condition and just always set the value to the generated path (making use of GNUInstallDirs to get it).

Perhaps the `resources/` part of the path isn't necessary? There are no files at that point, only under the `fonts` directory.

exjam:
Hmmm I guess it's kind of difficult really, because you can't really distinguish between a make and a make install. I guess maybe we either need to:

1. Get rid of all the resources - this means that we no longer ship CafeXXX fonts and instead ask people to put them in the correct MLC path (/vol/storage_mlc01/sys/title/0005001B/10042400/content), additionally we could add a decaf-qt UI based thing which will ask the user if they would like to download them automatically when they're not present. For other frontends which do not run through decaf-qt we would just have to log a warning. I also feel it is inevitable that we will need some "setup mlc/slc" UI in decaf-qt eventually  for setting up things like user accounts and other data that is expected to be found on the MLC / SLC paths. This leaves the DejaVu Sans Mono.ttf which we use for the imgui based debugger ui - this could easily be replaced with a system font if we had a list of appropriate fonts which are likely to be installed on Linux / Windows / MacOS.

2. Make the resources path a bit more intelligent, if it is default value of "resources" we could also check /usr/share/decaf-emu if resources are not found in $PWD/resources.

Tatsh:
I think 2 is the best option. Still would want to use CMake to get the path and define the macro since we need the prefix (could also just give the code the prefix).

If you want to not distribute the Deja Vu font, you could use the [Fontconfig API](https://linux.die.net/man/3/fcfontmatch) to find a good enough font for the purpose (this also allows the user customisation if they don't like your choice of font). I would pass 'monospace' to this API to find the user's default font for that style (I override all monospace with [Fira Code](https://github.com/tonsky/FiraCode) in my config). I think on most distros at this point, the matching font will be Deja Vu, but it could be [Ubuntu](https://design.ubuntu.com/font/) on Ubuntu distros. [Example](https://stackoverflow.com/a/14634033/374110)

brett19:
Or we could use bin2c to bundle the binary font data into the executable rather than having it be an external "dependency".  These seems to be the cleanest option really...

exjam:
I mean I personally prefer option 1 but I accept it would involve alot of work - that means we have 3 different opinions yay!

My reason for that being it is most accurate in the "emulation" sense - if theoretically we did LLE of the CafeOS kernel the we'd need to take this option (although i'm like 98.5% sure we'd never do LLE emulation of kernel). Also as I said there is other MLC / SLC setup which is required, so it could all be merged into one fancy System Setup Wizard on decaf-qt.

It would mean I could remove the loadResourcesFile function from cafe::kernel
https://github.com/decaf-emu/decaf-emu/blob/master/src/libdecaf/src/cafe/kernel/cafe_kernel_shareddata.cpp#L47-L108

Regarding bin2c, the font files are quite big so you'd be adding like 26.3MB to the .exe file.

exjam:
Maybe we should go with option 2 to resolve issues for the short term and have option 1 as a longer term goal.

Tatsh:
The above is now implemented in this PR.

exjam:
Merged with a few changes to fix Windows.

Issue #537, opened 2019-01-27
SQ_VTX_SEMANTIC endian issues causing shader compilation failures


QuarkTheAwesome:
Heyo! I've been working on an SDL2 port using wut, and it's been acting rather strange. I was hoping decaf might shed some light on the graphics issues, but ran into what looks like an endian handling bug.

Here's a log from running the [opentyrian.rpx](https://github.com/decaf-emu/decaf-emu/files/2799821/opentyrian.rpx.zip) on my Linux system, freshly compiled Decaf (commit 070e29c977c87ff05cce2fc29007d497a4b7cdcf): https://pastebin.com/Y6vDyXtF - skip to line 1993-ish. I'm using the OpenGL backend here since I can't get the Vulkan one to work right on my (AMD) machine. I've only really looked into the '_fs_out_0 redeclared' error, kinda hoping the pixel shader issue is something similar.

After much gdb-ing, it turns out that `sq_vtx_semantic.SEMANTIC_ID()` ([used here](https://github.com/decaf-emu/decaf-emu/blob/master/src/libgpu/src/opengl/opengl_shader.cpp#L1132)) always comes back as 0. Specifically:
```
(gdb) s
1134          if (id == 0xff) {
(gdb) p sq_vtx_semantic
$24 = {value = 0x1000000}
(gdb) p sq_vtx_semantic.SEMANTIC_ID()
$25 = 0x0
(gdb) p id
$26 = 0x0
```
From what I understand, sq_vtx_semantic is supposed to be 0x01 here, so that .SEMANTIC_ID can take the rightmost 8 bits. I tried printing `shader->regs.sq_vtx_semantic` on a real Wii U and yeah, the 01 is on the right, not the left. This makes me think it's an endian issue! I went and checked places it might be swapped:
- As soon as it's passed into [GX2SetVertexShader](https://github.com/decaf-emu/decaf-emu/blob/master/src/libdecaf/src/cafe/libraries/gx2/gx2_shaders.cpp#L55), all the elements in `shader->regs.sq_vtx_semantic` have not been swapped (01, ff etc. on the left, according to GDB)
- After the [`.value()`](https://github.com/decaf-emu/decaf-emu/blob/master/src/libdecaf/src/cafe/libraries/gx2/gx2_shaders.cpp#L67) call, the values in `sq_vtx_semantic` don't seem to have been swapped yet, the copy going on in be2_array never seems to be overridden with a swap function
[...]

Issue #538, opened 2019-01-28
Only use 1 Vulkan graphics queue


exjam:
AMD usually only has 1 graphics queue, so we must refactor how we handle rendering to support this.

Narann:
More information about queue family [here](https://www.reddit.com/r/vulkan/comments/46ckoc/queue_families/).

> From what I noticed:
> 
> * NVIDIA GPUs generally have 16 identical Queues (one queue family) that can do anything. I think these queues are multiplexed by the driver onto one actual hardware queue at drawcall boundaries.
> * AMD GPUs have 4+ queues of various families (one graphics/compute queue, 1-4 compute only queues and two transfer queues),
> * Intel GPUs **only have a single queue**.

exjam:
Fixed with b2d18768bd6ba3caf4184525f5aefaae50c0adbb

PR #544, opened 2019-02-09
fix bitfield byte orders on linux

snickerbockers:
Use a union for bitfields when DECAF_USE_STDLAYOUT_BITFIELD is defined

Previous behavior was to use a struct.  This was causing be2_array
instantiations to use be2_struct for be2_array_item_type instead of be2_val.
The upshot of this was that on Linux builds, byte order would be reversed for
bitfields wrapped in be2_array because be2_struct doesn't do endian reversal.
When DECAF_USE_STDLAYOUT_BITFIELD is not defined, bitfields were already unions,
which is why this behavior did not manifest on Windows.

This commit resolves several programs which were working on Windows but not
Linux and also addresses issue #537 on github.

### DIFF
--- a/src/common/bitfield.h
+++ b/src/common/bitfield.h
@@ -105,7 +105,7 @@ struct BitfieldHelper<BitfieldType, bool, Position, Bits>
 #else
 
 #define BITFIELD_BEG(Name, Type)                                                  \
-   struct Name                                                                \
+   union Name                                                                 \
    {                                                                          \
       using BitfieldType = Name;                                              \
       using StorageType = Type;                                               \


QuarkTheAwesome:
Can confirm that this fixes all the shader compiling issues I had - though my app still doesn't work right! :grin:

snickerbockers:
That might be a GL issue.  I have it working on my linux box when I use Vulkan, but on GL it's just a blank screen.  Do you know if it works on Windows OpenGL?

QuarkTheAwesome:
exjam sent me a screencap of it working under Vulkan on Windows, so yeah, it sounds like there's something up with GL (that or it's the message about ignoring early-Z)
Guess I'm back to waiting on #538 :upside_down_face: 
Still, thanks for looking into this!

exjam:
Great find - thanks.

PR #546, opened 2019-02-17
ios/mcp: if cos.xml doesn't have any permissions then grant full perm…

snickerbockers:
…issions


exjam:
85cc068d5d5ae63a8bf46ec445b3056e9ee0db73 thanks,

PR #553, opened 2019-05-07
vfs: fix makeFolder failure on GNU/Linux

snickerbockers:
SAVEInitSaveDir wants to create a directory hierarchy under usr/save/...
on Linux the create_directory call will fail because it only lets you create one directory at a time, so use create_directories instead.
This was not a problem on Windows because the /std:c++latest compiler option causes create_directory to behave the same as create_directories.

This commit fixes NES and SNES vc games which were working on Windows but not Linux.


exjam:
Silly Windows! Thanks

PR #561, opened 2019-06-24
libgpu/vulkan: Remove double negation

Narann:
Remove double negation in method arguments and checks.

### DIFF
--- a/src/libgpu/src/vulkan/vulkan_driver.h
+++ b/src/libgpu/src/vulkan/vulkan_driver.h
@@ -601,12 +601,12 @@ protected:
    void _invalidateSurface(SurfaceObject *surface, SurfaceSubRange range);
    void _barrierSurface(SurfaceObject *surface, ResourceUsage usage, vk::ImageLayout layout, SurfaceSubRange range);
    SurfaceObject * getSurface(const SurfaceDesc& info);
-   void transitionSurface(SurfaceObject *surface, ResourceUsage usage, vk::ImageLayout layout, SurfaceSubRange range, bool skipChangeCheck = false);
+   void transitionSurface(SurfaceObject *surface, ResourceUsage usage, vk::ImageLayout layout, SurfaceSubRange range, bool checkSurfaceChanges = true);
 
    SurfaceViewObject * _allocateSurfaceView(const SurfaceViewDesc& info);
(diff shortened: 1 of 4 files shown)


src/libgpu/src/vulkan/vulkan_driver.h
```diff
@@ -601,12 +601,12 @@ class Driver : public gpu::VulkanDriver, public Pm4Processor
    void _invalidateSurface(SurfaceObject *surface, SurfaceSubRange range);
    void _barrierSurface(SurfaceObject *surface, ResourceUsage usage, vk::ImageLayout layout, SurfaceSubRange range);
    SurfaceObject * getSurface(const SurfaceDesc& info);
-   void transitionSurface(SurfaceObject *surface, ResourceUsage usage, vk::ImageLayout layout, SurfaceSubRange range, bool skipChangeCheck = false);
+   void transitionSurface(SurfaceObject *surface, ResourceUsage usage, vk::ImageLayout layout, SurfaceSubRange range, bool doChangeCheck = true);
```
CodingKoopa: IMO the "do" here might be unecessary to have?

Narann: I don't know... `changeCheck` could be understands as "change the check".

CodingKoopa: What exactly is it doing then?

Narann: It checks the surface has changed.

CodingKoopa: `do` is fairly unconventional for bool names in my experience. Perhaps `checkSurfaceChanges`.

Narann: I agree, `checkSurfaceChanges` is better.

brett19:
Original implementation follows the 'principal of least surprise'.  In the case of skipping the surface checks, the default of a boolean is typical false, and in this case, the majority case involves performing the surface checks.  This is why we added a flag for skipping as opposed to a flag for checking.

Narann:
Thanks for the explanation Brett. I think false booleans makes code harder to read and logic flows hard to follow. Maybe it’s just me.

PR #567, opened 2019-09-22
Set up CI with Azure Pipelines

emulationchannel:
[skip ci]


exjam:
I am intending on moving decaf to Github Actions rather than using travis/appveyor/azure - I was actually working on this yesterday but Conan has an issue with their CDN atm which is stopping me from finished it right now.

emulationchannel:
Ok Sorry for the inconvenience of this PR Decaf-Emu is not spam if you think or think because many think I spam

PR #573, opened 2019-10-19
cmake: Set CMP0072 policy

dracc:
This sets a policy deciding what version of available OpenGL libraries to use.
Silences a warning that pops up when running cmake.


exjam:
Does this work on older CMake versions ? I see the new policy was added in CMake 3.11 but not sure if it is an error on cmake versions before that.

dracc:
That's a good question. I personally don't have the means to test it, would be great if someone else could!

dracc:
I added an `if(POLICY)` check, should spare us from new errors.

PR #576, opened 2019-11-08
fix linux build

snickerbockers:
the case-sensitivity of the include-statement didn't match the case-sensitivity
of the file.


exjam:
I should add Qt to the Linux CI build..!

PR #577, opened 2019-11-23
Added command line options to decaf-qt

SteveLeafo:
Small change to allow games to be started from the command line in a similar fashion to the sdl application.

### DIFF
--- a/src/decaf-qt/src/main.cpp
+++ b/src/decaf-qt/src/main.cpp
@@ -12,8 +12,17 @@
 #include <SDL2/SDL.h>
 
 #include <QApplication>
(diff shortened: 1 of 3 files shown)


exjam:
decaf-qt should use excmd so we can inherit the same command line options provided by other decaf applications: https://github.com/decaf-emu/decaf-emu/blob/master/src/libconfig/src/loader_excmd.cpp

for reference you can see decaf-sdl: 
https://github.com/decaf-emu/decaf-emu/blob/master/src/decaf-sdl/main.cpp

brett19:
Closing this PR in favor of issue #592

PR #583, opened 2019-12-16
libdecaf: Fix lib name in default trace filters

Kinnay:
My trace filters didn't work until I noticed that I have to add the .rpl extension to the library names. Would have prevented some confusion if the default trace filters were correct.


brett19:
Haha.  Woops...  I imagine both should probably work to be fair.

PR #584, opened 2019-12-16
vfs: handle ENOENT on posix systems

Kinnay:
For some reason, std::filesystem::status sets the std::error_code to ENOENT, instead of returning the appropriate std::filesystem::file_status. Decaf only checks the error code on windows systems. On other systems, it always returns Error::GenericError, which is translated to FSAStatus::MediaError. Many games are crashing because of this.

### DIFF
--- a/src/libdecaf/src/vfs/vfs_host_device.cpp
+++ b/src/libdecaf/src/vfs/vfs_host_device.cpp
@@ -8,6 +8,9 @@
 #ifdef PLATFORM_WINDOWS
 #include <winerror.h>
 #endif
+#ifdef PLATFORM_POSIX
+#include <errno.h>
+#endif
 
 namespace vfs
 {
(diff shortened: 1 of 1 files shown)


brett19:
I believe the real solution here is to actually use the return value of the status call to deduce the state of the file first, and then progress to using the error_code after that.  👎

Kinnay:
I guess that makes sense. I assumed that the return value is always file_type::none if the error code is non-zero because that's the only case where cppreference explicitly mentions that it sets the error code, but actually it correctly returns file_type::not_found in this case.

However, what are you going to do with the functions that don't return a file_status? For example, what if HostDevice::rename is called on a file that doesn't exist?

exjam:
This is strange indeed, according to https://en.cppreference.com/w/cpp/filesystem/status

`If p does not exist, returns file_status(file_type::not_found)`

So you would expect it to not error. I wonder if that can be considered a bug in the standard library implementation.

You are right - we definitely need to add support for translateError for non-Windows systems.

exjam:
Looks like both libstdc++ and libcxx use generic category with errno - so I think you have taken the right approach here.

https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/src/filesystem/ops.cc#L1188-L1210

https://github.com/llvm-mirror/libcxx/blob/7c3769df62c0b3820130aa868397a80a042e0232/src/filesystem/filesystem_common.h#L109-L112

exjam:
I went ahead and did this using std::errc which will work for both Windows and Posix errno stuff: https://github.com/decaf-emu/decaf-emu/commit/b26df065b42adf326fb0ce666b5f85864bfe4e5f

Thank you for finding the issue and for your work towards finding a fix though :)

PR #585, opened 2019-12-17
cafe/coreinit: Fix buffer size in OSPanic

Kinnay:
A lot of garbage is printed right now.


exjam:
Thanks for this, I did notice it before in logs - was very annoying!

brett19:
Hehe.. Oops!  Thanks for fixing this Kinnay, was my fault.

PR #594, opened 2019-12-24
ios/mcp: Update title id in mcpPrepareTitle52 if it is DefaultTitleId.

Kinnay:
I saw the following line in the log: `[08:49:51.880345 info:decaf] No title update found at /vol/storage_mlc01/usr/title/0005000e/fffffffd`

I don't think that's the intended behavior.


exjam:
Nice catch, thanks.

PR #607, opened 2020-10-09
libgpu: check return codes, not exceptions, for noexcept vulkan funcs

QuarkTheAwesome:
Around January, vulkan.hpp changed some key functions to be `noexcept`, including `vk::Device::acquireNextImageKHR` and `vk::Queue::presentKHR` ([commit](https://github.com/KhronosGroup/Vulkan-Hpp/commit/e49f02013a5cf32c14e2f912e16e4fa181c8223a)). Decaf checked for swapchain out-of-date conditions with try/catch, which doesn't work on Vulkan SDK versions with this header.

Tested OK on Vulkan SDK 1.2.154.0, linux, mesa 20.2 (selfbuild).


exjam:
Thanks! vulkan.hpp = 💩

PR #608, opened 2020-11-16
Replace deprecated Vulkan validation

curtisy1:
VK_LAYER_LUNARG_standard_validation was deprecated in Vulkan 1.1.106.0 and replaced by  VK_LAYER_KHRONOS_validation. The CI already uses Vulkan 1.2x but when running locally with up to date packages, validation errors occur since the standard_validation is not included in latest Vulkan versions anymore (at least on my machine).

Since VK_LAYER_KHRONOS_validation was introduced after 1.1.92.1, the minimum required version also had to be bumped. I take it this is not an issue though since 1.1.108 is more than one and a half years old by now

Also includes CI fixes for the recent [set-env deprecation](https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/) and [XCode11 deprecation](https://github.com/actions/virtual-environments/issues/1688)

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -240,7 +240,7 @@ jobs:
       shell: cmd
       run: |
         cd build/install
-        $QTDIR/bin/windeployqt.exe --release --no-translations --no-angle --no-opengl-sw decaf-qt.exe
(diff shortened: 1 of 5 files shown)


exjam:
Nice one thanks

PR #614, opened 2021-04-13
Fix QObject::connect: No such slot MainWindow::toggleFullScreen()

i-can-not-program:
Fix
QObject::connect: No such slot MainWindow::toggleFullScreen()
QObject::connect:  (sender name:   'actionViewFullScreen')
QObject::connect:  (receiver name: 'MainWindow')


exjam:
Thanks

PR #617, opened 2021-07-17
latte-assembler: Add additional CfInstProperties

GaryOderNichts:
Makes it possible to use `CF_CONST`, `FAIL_JUMP_ADDR` and `PASS_JUMP_ADDR` on control flow instructions.
This allows you to assemble loops with latte-assembler.


exjam:
Thanks !

PR #623, opened 2021-12-18
Audio fixes for 32kHz games

QuarkTheAwesome:
One of the codepaths through the final mix callback would zero 2/3 of the samples before outputting them, which is exciting.
Two patches - one to not zero the samples, and one to let the final mix callback see all the samples, not just 2/3.

Restores audio to Mario Kart 8 and fixes corruption in PUYOPUYOTETRIS. I still get some crunchiness on the audio but that could be my OS not liking when the framerate drops; haven't chased it.

Needs testing on Windows


exjam:
Thanks!

PR #624, opened 2021-12-19
cafe/sndcore2: Fix ADPCM decoding

QuarkTheAwesome:
Looks like this code came from [BrawlLib](https://github.com/libertyernie/brawltools/blob/master/BrawlLib/Wii/Audio/ADPCMState.cs#L68-L71); but has its condition flipped?
Fix makes Mario Kart 8 sound much better.


exjam:
Thanks!

PR #625, opened 2021-12-22
libdecaf: put usercontext in the iosu thread structure

DacoTaco:
ive been reverse engineering IOS on the wii for a while now and i recently got through its syscalls/thread scheduler.
ive been using the decaf code for a lot of stuff since it shares a lot of code and ideas.

on the wii, the IOS kernel sets the supervisor & undf stackpointer to allocated space within the running's thread structure.
when saving the user thread state, it saves it to the thread structure and then switches to system mode (which shares stackpointer with the thread).

Today i decrypted and verified the same behavior in IOSU, so this fixes that bit.
from what i saw all of this is unused within decaf, but from a documentation standpoint, its more accurate :)

from IOSU's thread schedule : 
image

and then in the undefined handler : 
image

from IOS (ignore the very very ugly code, im yolo'ing all of it without deciding on code guidelines)
https://github.com/DacoTaco/StarStruck/blob/master/kernel/source/interrupt/threads.c#L148-L169

### DIFF
--- a/src/libdecaf/src/ios/ios_stackobject.h
+++ b/src/libdecaf/src/ios/ios_stackobject.h
@@ -20,9 +20,9 @@ public:
    StackObject()
    {
       auto thread = kernel::internal::getCurrentThread();
-      auto ptr = phys_cast<uint8_t *>(thread->stackPointer) - AlignedSize;
+      auto ptr = phys_cast<uint8_t *>(thread->userContext.stackPointer) - AlignedSize;
       phys_ptr<Type>::mAddress = phys_cast<phys_addr>(ptr);
-      thread->stackPointer = ptr;
+      thread->userContext.stackPointer = ptr;
 
       std::uninitialized_default_construct_n(phys_ptr<Type>::get(),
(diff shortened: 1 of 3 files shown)


exjam:
Pushed with a minor fix as 9b4a73129f6c4dce919919bc36dce30550561d43.

Thanks !

DacoTaco:
euh, this merged correctly? i noticed the ci builds failed? did it run on my old commit (before i removed the ifdef's ? )

exjam:
Your change had two problems which I fixed before pushing:

`be2_phys_ptr<ContextLLE> userContext;`
instead of
`be2_struct<ContextLLE> userContext;`

and

`be2_val<uint32_t> stackPointer;`
instead of
`be2_phys_ptr<void> stackPointer;`

I amended your commit to fix that and pushed it, that's why in my comment I said "pushed with a **minor fix**".

DacoTaco:
ah ok, sorry ^^;
next time ill get all the required tools and build it myself first as well. i didn't expect such issues...

PR #630, opened 2022-01-14
Clarify artifact downloads

CodingKoopa:
Yet Another Readme PR, sorry. This changes the readme instructions to filter out the action runs to only include successful builds on `master`. The section mentions all 3 OSes now, with Linux's state being changed to reflect it now having binaries.

GitHub does not provide a way for users to download the latest artifact from one link. This appears to be achievable via https://nightly.link/, but that would be outside of the scope of this PR.


exjam:
Sensible change, thanks !

PR #631, opened 2022-01-18
CI: Create releases

qurious-pixel:


.github/workflows/ccpp.yml
```diff
@@ -205,3 +205,29 @@ jobs:
...
+          mkdir artifacts/
+          files=$(find . -name "decaf-emu-*" ! -iname "*.zip")
+          for f in $files; do
+            echo "Compressing $f"
+            (cd $(basename $f) && zip -r ../artifacts/$(basename $f).zip *)
+          done
+          ls -al artifacts/
+          wget -c https://github.com/tcnksm/ghr/releases/download/v0.14.0/ghr_v0.14.0_linux_amd64.tar.gz
+          tar xfv ghr_v0.14.0_linux_amd64.tar.gz
+          ghr_v0.14.0_linux_amd64/ghr -u qurious-pixel -r decaf-emu -recreate -n 'decaf-emu CI builds' -b "Corresponding commit: ${{ github.sha }}" continuous artifacts/
```
CodingKoopa, line 231: How does this authenticate? Upstream seems to use a token specified via a Git configuration, or via `-t`, and in any case the team should be directed to create the access token.

On a similar note, your personal GitHub account probably shouldn't be here.

.github/workflows/ccpp.yml
```diff
@@ -205,3 +205,29 @@ jobs:
...
+        shell: bash
+        run: |
+          mkdir artifacts/
+          files=$(find . -name "decaf-emu-*" ! -iname "*.zip")
+          for f in $files; do
+            echo "Compressing $f"
+            (cd $(basename $f) && zip -r ../artifacts/$(basename $f).zip *)
+          done
+          ls -al artifacts/
+          wget -c https://github.com/tcnksm/ghr/releases/download/v0.14.0/ghr_v0.14.0_linux_amd64.tar.gz
```
CodingKoopa, line 229: Is hardcoding this to a fixed release ideal?

CodingKoopa (commented):
Thanks! I left a couple of thoughts.

.github/workflows/ccpp.yml
qurious-pixel: This reverted when I fixed the spacing, correcting now.
The token comes from the user with write permissions to the repository. 
A separate token can be specified with `-t TOKEN`

.github/workflows/ccpp.yml
qurious-pixel: Latest repo can be used, but specifying does add some stability. Known good version.

exjam:
Manually merged this, let's see if it works :). Thanks !

PR #632, opened 2022-01-21
cafe/gx2: Sanitise frontFace in GX2InitPolygonControlReg

QuarkTheAwesome:
Hyrule Warriors calls with garbage values, the original gx2 code does an &1 
to solve that.

Allows Hyrule warriors to get a little further, before a rather mystifying stack/heap corruption issue appears instead.


exjam:
Nice one thanks !

PR #634, opened 2022-01-23
cmake: Add hint for ffmpeg headers on Linux

QuarkTheAwesome:
Apparently some distros put it here instead - maybe this should use pkg-config eventually? Though I guess you have other plans with vcpkg.

I usually just disabled ffmpeg but it's a hard dependency now.


exjam:
I should probably make it optional again, I made it hard dependency before because I was thinking every platform could use vcpkg but that was too wishful thinking :).

exjam:
Merged anyway as no harm to adding hints, but I have also restored the DECAF_FFMPEG option.

PR #637, opened 2022-01-25
cafe/sndcore2: Add AXVoiceLpf

QuarkTheAwesome:
An implementation of the AX LPF. The only sound I could actually find which uses it is the back button on the NSMBU title screen, so I didn't have much opportunity to test it thoroughly, but the code agrees with both RE and [the maths](https://en.wikipedia.org/wiki/Low-pass_filter#Simple_infinite_impulse_response_filter).

Things to note:
- Cafe does all its effects and mixing internally with floats, so I may have messed up somewhere in the conversion to fixed-point maths.
- I put `sfixed_1_0_15_t` in the Cafe-facing HLE API, maybe that's not desirable, I dunno.

At least one other test case would be appreciated if anyone has one.

### DIFF
--- a/src/libdecaf/src/cafe/libraries/sndcore2/sndcore2_device.cpp
+++ b/src/libdecaf/src/cafe/libraries/sndcore2/sndcore2_device.cpp
@@ -296,6 +296,24 @@ sampleVoice(virt_ptr<AXVoice> voice,
 }
 
 void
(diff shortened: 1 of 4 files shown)


QuarkTheAwesome:
Other than the MSVC issue, I suspect I did in fact screw up when converting from fixed to floating math, which means this implementation is straight-up wrong.
Bear with me!

PR #638, opened 2022-01-26
cafe/h264: Update ffmpeg api and fix flushing behaviour.

QuarkTheAwesome:
Updates to the newer `av_packet_alloc` API, which removes some deprecation warnings. Also changes the flushing behavior - more details in the commit text - which fixes video playback in Mario Kart 8.


exjam:
Nice one thanks! I did notice yesterday that alot of videos seemed to not be playing correct but didn't look into it.

PR #639, opened 2022-01-26
cafe/sndcore2: Implement voice ADSR

QuarkTheAwesome:
Adds "ADSR" (actually just per-voice volume scaling) emulation, which significantly improves audio in NSMBU and Mario Kart 8 (the engine sounds specifically). Not much to say, the code is self-explanatory, and unlike my attempts to implement the filters, I know this is correct~


exjam:
Great job, the audio is much better with this and your other fix! Thanks.

PR #641, opened 2022-01-30
latte-assembler: Support POP_CNT and ADDR for control flow instructions

GaryOderNichts:
`ADDR` is needed to specify the target address for `JUMP` instructions.
`POP_CNT` can be used on control flow instructions to pop the pixel state from the stack.


exjam:
Thanks !

PR #642, opened 2022-02-18
Unbundle libs

mazes-80:
As said in the last bug report opened, I submit some very small tweaks to the build system.

I allows using some system libraries instead of bundled one when requirement are found


mazes-80:
Yes totally agree. I'll update PR

mazes-80:
Ouch that hurt, I thought it was done. Sorry
Fortunately I  still have the local clone. I just re-pushed the branch and this will make me think about steps to finish

PR #651, opened 2023-10-20
ios/kernel: Improve `IOS_HeapRealloc`  

GaryOderNichts:
This fixes two small mismatches I noticed when comparing against the IOSU implementation.


exjam:
Thanks !

